PR #7, opened 2018-04-23
Added help x command that lists all possible sound effects

DzikowskiW:


Deseteral:
This PR is contributing to the old version of backend. It is outdated.

Issue #5, opened 2018-04-20
Playing the song should increment its play count


Deseteral:
When a song is played, TV dashboard should send song played event with its information to the backend service so we can bump its play count.

AC:
- Playing the song increments its play count

PR #9, opened 2018-04-24
New features

Deseteral:
- application will try to fetch video title and automagically add to database unknown videos
- `search` bumps play count
- `queue` bumps play count
- new command: `xlist` - sends list of xsounds
- xsounds now have a play count


Deseteral:
Fixes #5

PR #39, opened 2018-10-25
Added relative volume commands

### LINKED ISSUE #32: Relative volume control
Allows users to send commands with relative volume.
Ex.
`/fm vol +10` will increate volume by 10%
`/fm vol -23` will decrease volume by 23%

This change doesn't break absolute volume control so `/fm vol 65` will set volume to exactly 65%.

DzikowskiW:
Fixes #32

### DIFF
--- a/server/commands/VolumeCommand.ts
+++ b/server/commands/VolumeCommand.ts
@@ -2,24 +2,16 @@ import CommandDefinition from '../domain/CommandDefinition';
 import IoConnection from '../clients/IoConnection';
 import VolumeEventMessage from "../domain/event-messages/VolumeEventMessage";
 
-const volChangeTypes: { [s: string]: string; } = {
-  '+': 'increase',
-  '-': 'decrease',
-  default: 'set',
-};
-
-function getVolChangeType(value: string):string {
-  return volChangeTypes[value.charAt(0)] || volChangeTypes.default;
-}
-
-function getFeedbackMsg(eventMsg:VolumeEventMessage) {
-  if (volChangeTypes['+'] === eventMsg.changeType) {
-    return `Zwiększono głośność o "${eventMsg.volume}"`;
+function getFeedbackMessage(eventMsg:VolumeEventMessage) {
+  const { volume } = eventMsg;
+  switch (volume[0]) {
+    case '+':
+      return `Zwiększono głośność o "${volume.substring(1)}"`;
+    case '-':
+      return `Zmniejszono głośność o "${volume.substring(1)}"`;
+    default:
+      return `Ustawiono głośność na "${volume}"`;
   }
-  if (volChangeTypes['-'] === eventMsg.changeType) {
-    return `Zmniejszono głośność o "${eventMsg.volume}"`;
-  }
-  return `Ustawiono głośność na "${eventMsg.volume}"`;
 }
 
 async function volume(value: string) : Promise<string> {
@@ -29,13 +21,13 @@ async function volume(value: string) : Promise<string> {
     return `Nieprawidłowa głośność ${value}, podaj liczbę z zakresu 0-100, lub deltę +/-liczba`;
   }
 
-  const eventMessage: VolumeEventMessage = { 
-    volume: parsedInt,
-    changeType: getVolChangeType(value),
+  const eventMessage: VolumeEventMessage = {
+    volume: value,
   };
+
   IoConnection.broadcast('volume', eventMessage);
 
-  return getFeedbackMsg(eventMessage);
+  return getFeedbackMessage(eventMessage);
 }
 
 const commandDefinition: CommandDefinition = {
--- a/server/commands/VolumeCommand.ts
+++ b/server/commands/VolumeCommand.ts
@@ -5,7 +5,7 @@ import VolumeEventMessage from "../domain/event-messages/VolumeEventMessage";
 const volChangeTypes: { [s: string]: string; } = {
   '+': 'increase',
   '-': 'decrease',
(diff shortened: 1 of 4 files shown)


server/commands/VolumeCommand.ts
```diff
@@ -2,17 +2,40 @@ import CommandDefinition from '../domain/CommandDefinition';
 import IoConnection from '../clients/IoConnection';
 import VolumeEventMessage from "../domain/event-messages/VolumeEventMessage";
 
+const volChangeTypes: { [s: string]: string; } = {
+  '+': 'increase',
+  '-': 'decrease',
+  default: 'setter',
```
Deseteral: `increase` and `decrease` are both verbs. `setter` is not.
```suggestion
  default: 'set',
```

server/commands/VolumeCommand.ts
```diff
@@ -2,17 +2,40 @@ import CommandDefinition from '../domain/CommandDefinition';
 import IoConnection from '../clients/IoConnection';
 import VolumeEventMessage from "../domain/event-messages/VolumeEventMessage";
 
+const volChangeTypes: { [s: string]: string; } = {
```
Deseteral: Prefer volume over vol.
```suggestion
const volumeChangeTypes: { [s: string]: string; } = {
```

server/commands/VolumeCommand.ts
```diff
@@ -2,17 +2,40 @@ import CommandDefinition from '../domain/CommandDefinition';
...
+  '+': 'increase',
+  '-': 'decrease',
+  default: 'setter',
+};
+
+function getVolChangeType(value: string):string {
+  return volChangeTypes[value.charAt(0)] || volChangeTypes.default;
+}
+
+function getFeedbackMsg(eventMsg:VolumeEventMessage) {
```
Deseteral: Prefer Message over Msg.
```suggestion
function getFeedbackMessage(eventMesssage: VolumeEventMessage) {
```

src/services/serverPush.ts
```diff
@@ -59,7 +60,13 @@ function handleX(xsound: IXMessage) {
 }
 
 function handleVolume(xvolume: IVolumeMessage) {
```
Deseteral: Why is this argument called `xvolume`? Shouldn't it be just called `volume`?

src/services/youtubePlayer.js
```diff
@@ -53,7 +53,19 @@ function changeVolume(val) {
...
 }
 
-function triggerOnVideoChange(){
+async function increaseVolume(val) {
+  const oldVolume = await player.getVolume();
+  const newVolume = Math.min(oldVolume + val, 100);
+  return player.setVolume(newVolume);
+}
+
+async function decreaseVolume(val) {
```
Deseteral: Prefer value over val.
```suggestion
async function decreaseVolume(value) {
```

server/commands/VolumeCommand.ts
```diff
@@ -2,17 +2,40 @@ import CommandDefinition from '../domain/CommandDefinition';
 import IoConnection from '../clients/IoConnection';
 import VolumeEventMessage from "../domain/event-messages/VolumeEventMessage";
 
+const volChangeTypes: { [s: string]: string; } = {
+  '+': 'increase',
+  '-': 'decrease',
+  default: 'set',
```
Deseteral: Both this object and `changeType` field inside `VolumeEventMessage` should be the same enum type.
https://www.typescriptlang.org/docs/handbook/enums.html

server/commands/VolumeCommand.ts
```diff
@@ -2,17 +2,40 @@ import CommandDefinition from '../domain/CommandDefinition';
...
+};
+
+function getVolChangeType(value: string):string {
+  return volChangeTypes[value.charAt(0)] || volChangeTypes.default;
```
Deseteral: I'm curious, why `charAt(0)` instead of `[0]`?

server/commands/VolumeCommand.ts
```diff
@@ -2,17 +2,40 @@ import CommandDefinition from '../domain/CommandDefinition';
...
+}
+
+function getFeedbackMsg(eventMsg:VolumeEventMessage) {
+  if (volChangeTypes['+'] === eventMsg.changeType) {
+    return `Zwiększono głośność o "${eventMsg.volume}"`;
+  }
+  if (volChangeTypes['-'] === eventMsg.changeType) {
+    return `Zmniejszono głośność o "${eventMsg.volume}"`;
+  }
+  return `Ustawiono głośność na "${eventMsg.volume}"`;
```
Deseteral: Probably will look better with switch statement instead of else-if chain.

server/commands/VolumeCommand.ts
```diff
@@ -2,17 +2,40 @@ import CommandDefinition from '../domain/CommandDefinition';
...
+}
+
 async function volume(value: string) : Promise<string> {
-  const parsedInt: number = parseInt(value, 10);
+  const parsedInt: number = Math.abs(parseInt(value, 10));
```
Deseteral: Wouldn't it be simpler to pass positive or negative value and have two modes: relative and absolute?

Now you have three modes: setting, increasing and decreasing. Increasing and decreasing are very similar and come down to just adding positive or negative value to current volume.

This looks like simpler, less-code-required solution. Am I missing something?

DzikowskiW: done

Deseteral:
Closing this PR due to upcoming rewrite of the project.

PR #40, opened 2018-10-25
Added command to queue a song at the start of a queue

DzikowskiW:


server/commands/QueueNextCommand.ts
```diff
@@ -0,0 +1,46 @@
+import CommandDefinition from '../domain/CommandDefinition';
+import IoConnection from '../clients/IoConnection';
+import Song from '../domain/Song';
+import SongRepository from '../repositories/SongRepository';
+import QueueEventMessage from '../domain/event-messages/QueueEventMessage';
+import FetchVideoTitle from '../helpers/FetchVideoTitle';
+import SongService from '../services/SongService';
+
+async function createSongFromId(youtubeId: string) : Promise<Song> {
```
Deseteral: This looks like something that could be extracted for additional reusability (same thing is used inside queue command service).

There is helpers module that looks like a good place for this (it looks like a good place for anything really...)

server/commands/QueueNextCommand.ts
```diff
@@ -0,0 +1,46 @@
...
+    timesPlayed: 0,
+  };
+
+  return song;
+}
+
+async function queueNext(songName: string) : Promise<string> {
+  const song: Song = await SongRepository.getByName(songName);
+  const potentialYoutubeId = songName.split(' ')[0];
+  const actualSong = song
```
Deseteral: This could be simplified to simple or statement I think?
```suggestion
  const actualSong = song || await createSongFromId(potentialYoutubeId);
```

server/commands/QueueNextCommand.ts
```diff
@@ -0,0 +1,46 @@
...
+    ? song
+    : await createSongFromId(potentialYoutubeId);
+
+  actualSong.playNext = true;
+  const eventMessage: QueueEventMessage = { song: actualSong, playNext: true };
```
Deseteral: Why are you setting `playNext` on both `actualSong` and `eventMessage`?

server/domain/Song.ts
```diff
@@ -5,6 +5,7 @@ interface Song {
   trimStartSeconds: (number | null);
   trimEndSeconds: (number | null);
   timesPlayed: number;
+  playNext?: boolean;
```
Deseteral: This should be required field

server/domain/event-messages/QueueEventMessage.ts
```diff
@@ -2,6 +2,7 @@ import Song from '../Song';
 
 interface QueueEventMessage {
   song: Song;
+  playNext?: boolean;
```
Deseteral: This should be required field

src/services/serverPush.ts
```diff
@@ -14,6 +14,7 @@ interface ISong {
   trimStartSeconds: (number | null);
   trimEndSeconds: (number | null);
   timesPlayed: number;
+  playNext?: boolean;
```
Deseteral: This should be required field. In the future it would be awesome if we could share type definitions between frontend and backend. This is one of those places where TypeScript shines 😄

Deseteral:
Closing this PR due to upcoming rewrite of the project.

PR #81, opened 2020-10-15
add speed command

gorzalczany:


packages/fm-player/src/services/event-stream-client.ts
```diff
@@ -46,6 +47,11 @@ function connect() {
         YouTubePlayerService.playNextSong();
         break;
 
+      case 'SpeedEvent': {
+        const nextSpeed = eventData.nextSpeed as SpeedControl;
```
gorzalczany, line 51: ⚠️

packages/service/src/event-stream/model/events.ts
```diff
@@ -51,6 +56,7 @@ type EventData =
   | PauseEvent
   | ResumeEvent
   | SkipEvent
+  | SpeedEvent
```
gorzalczany, line 59: żeby uspójnić to 
  SpeedEvent -> ChangeSpeedEvent 
albo 
  ChangeVolumeEvent -> VolumeEvent. 
WDYT?

packages/service/src/event-stream/model/events.ts
```diff
@@ -63,5 +71,6 @@ export {
   PauseEvent,
   ResumeEvent,
   SkipEvent,
+  ChangeSpeedEvent as SpeedEvent,
```
Deseteral, line 74:
```suggestion
  ChangeSpeedEvent,
```

PR #96, opened 2020-10-28
added conditions to search query (only embeddable videos, no safe sea…

mictyd:
…rch)

### DIFF
--- a/packages/fm-player/src/services/youtube-player-service.ts
+++ b/packages/fm-player/src/services/youtube-player-service.ts
@@ -76,6 +76,7 @@ function initialize(playerContainerDomId: string) {
     const state = PlayerStateService.getState();
 
     player.setVolume(state.volume);
+    player.setPlaybackQuality('highres');
 
     if (state.currentlyPlaying) {
       const { song, time } = state.currentlyPlaying;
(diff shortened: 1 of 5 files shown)


packages/fm-player/src/services/youtube-player-service.ts
```diff
@@ -76,6 +76,7 @@ function initialize(playerContainerDomId: string) {
     const state = PlayerStateService.getState();
 
     player.setVolume(state.volume);
+    player.setPlaybackQuality('highres');
```
gorzalczany, line 79: The getPlaybackQuality, setPlaybackQuality, and getAvailableQualityLevels functions are no longer supported. In particular, calls to setPlaybackQuality will be no-op functions, meaning they will not actually have any impact on the viewer's playback experience.

packages/service/src/youtube/youtube-data-client.ts
```diff
@@ -47,6 +47,9 @@ class YouTubeDataClient {
     url.searchParams.set('q', phrase);
     url.searchParams.set('maxResults', maxResults.toString());
     url.searchParams.set('part', 'snippet');
+    url.searchParams.set('type', 'video');
+    url.searchParams.set('safeSearch', 'none');
+    url.searchParams.set('videoEmbeddable', 'true');
```
gorzalczany, line 52: 👏

gorzalczany (changes requested):

packages/fm-player/src/services/youtube-player-service.ts
gorzalczany: Ale jeśli działa to niech zostanie

gorzalczany (approved):

Deseteral (approved):

PR #99, opened 2020-10-30
daily limit queue next 

lukaszbojarski:
daily limit of using queue --next command


Deseteral:
We never cared about limits in LebkuchenFM and I don't see that changing anytime soon

PR #119, opened 2021-10-04
Tags!

### LINKED ISSUE #90: Tagging system for XSounds
- Allow users to add and remove tags from XSounds using commands
- Add command to list sounds with specific tag assigned to them

Deseteral:
This PR adds tagging system for XSounds and introduces five new commands to manage them:
- `/fm tag-list`
- `/fm tag-search`
- `/fm tag-show`
- `/fm tag-add`
- `/fm tag-remove`

I've also fixed some validation issues for `/fm x command`.

Closes #90.

### DIFF
--- a/packages/service/src/domain/commands/processors/tag-add-command.ts
+++ b/packages/service/src/domain/commands/processors/tag-add-command.ts
@@ -22,7 +22,8 @@ const tagAddCommandDefinition: CommandDefinition = {
   processor: tagAddCommandProcessor,
   helpMessage: 'Dodaje tag do podanego dźwięku',
   helpUsages: [
-    '<tag-name> <sound name>',
+    '<tag-name>|<sound name>',
+    'fun stuff|airhorn',
   ],
 };
 
--- a/packages/service/src/domain/commands/processors/tag-add-command.ts
+++ b/packages/service/src/domain/commands/processors/tag-add-command.ts
@@ -1,0 +1,29 @@
+import Command from '../model/command';
+import CommandProcessingResponse, { makeSingleTextProcessingResponse } from '../model/command-processing-response';
+import CommandDefinition from '../model/command-definition';
+import XSoundsService from '../../x-sounds/x-sounds-service';
+
+async function tagAddCommandProcessor(command: Command): Promise<CommandProcessingResponse> {
+  const commandArgs = command.getArgsByDelimiter('|');
+
+  if (commandArgs.length < 2) {
+    throw new Error('Zbyt mała liczba argumentów');
+  }
+
+  const [tagName, soundName] = commandArgs;
+
+  await XSoundsService.instance.addTag(soundName, tagName);
+  return makeSingleTextProcessingResponse(`Dodano tag "${tagName}" do dźwięku ${soundName}`, false);
+}
+
+const tagAddCommandDefinition: CommandDefinition = {
+  key: 'tag-add',
+  shortKey: 'ta',
+  processor: tagAddCommandProcessor,
+  helpMessage: 'Dodaje tag do podanego dźwięku',
+  helpUsages: [
+    '<tag-name> <sound name>',
+  ],
+};
+
+export default tagAddCommandDefinition;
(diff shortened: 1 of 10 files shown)


packages/service/src/domain/commands/processors/tag-search-command.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const tagSearchCommandDefinition: CommandDefinition = {
+  key: 'tag-search',
+  shortKey: 'ts',
+  processor: tagSearchCommandProcessor,
+  helpMessage: 'Wyszukuje dźwięki z danym tagiem',
+  helpUsages: [
+    '<tag name>',
+    'fun stuff',
```
gorzalczany, line 31: I thought that tags with space are disallowed

Deseteral: I went above and beyond to make spaces possible.


(I've copy-pasted code from `addx` command)

packages/service/src/domain/commands/processors/tag-add-command.ts
```diff
@@ -0,0 +1,30 @@
...
+    throw new Error('Zbyt mała liczba argumentów');
+  }
+
+  const [tagName, soundName] = commandArgs;
+
+  await XSoundsService.instance.addTag(soundName, tagName);
+  return makeSingleTextProcessingResponse(`Dodano tag "${tagName}" do dźwięku ${soundName}`, false);
+}
+
+const tagAddCommandDefinition: CommandDefinition = {
```
gorzalczany, line 19: `tag` seems like complicated functionality
What do you think about adding `tag-help` which displays help for all `tag-xyz` commands?

packages/service/src/domain/commands/processors/tag-remove-command.ts
```diff
@@ -0,0 +1,29 @@
+import Command from '../model/command';
+import CommandProcessingResponse, { makeSingleTextProcessingResponse } from '../model/command-processing-response';
+import CommandDefinition from '../model/command-definition';
+import XSoundsService from '../../x-sounds/x-sounds-service';
+
+async function tagRemoveCommandProcessor(command: Command): Promise<CommandProcessingResponse> {
+  const commandArgs = command.getArgsByDelimiter('|');
+
+  if (commandArgs.length < 2) {
+    throw new Error('Zbyt mała liczba argumentów');
```
gorzalczany, line 10: Maybe lets provide correct usage in response instead of returning generic message

packages/service/src/domain/commands/processors/tag-add-command.ts
gorzalczany: we could even provide `tag` as shortKey for it

Deseteral: So like `/fm help` but with `tag-*` filter? Or some other information that is not included in `/fm help`?

packages/service/src/domain/commands/processors/tag-remove-command.ts
Deseteral: I thought about it, but this seems like feature that all commands could use. This is beyond the scope of this PR in my opinion.

packages/service/src/domain/commands/processors/tag-add-command.ts
gorzalczany: like help with tag filter 👍🏻

packages/service/src/domain/commands/processors/tag-remove-command.ts
gorzalczany: We could adres it locally for tag-* in this PR and "globally" as separate issue. But I will not insist.

Deseteral: You're right - I'll change that

packages/service/src/domain/commands/processors/tag-add-command.ts
Deseteral: I'm not sure about that. Adding new command that does basically what `/fm help` seems unnecessary 🤔 .
Maybe we can test how `/fm help` looks with new commands on Slack and then decide?

PR #134, opened 2021-10-15
handle search phrase in random

gorzalczany:
Improve random command


Deseteral:
Seems fine for now, but it would be good to write some tests for that command.

PR #146, opened 2021-10-29
improve soundboard filter

gorzalczany:


packages/fm-player/src/components/SoundBoard.tsx
```diff
@@ -5,7 +5,9 @@ import SoundButton from './SoundButton';
 import Search from './Search/Search';
 
 function soundMatchesPhrase(sound: XSound, phrase: string) {
-  return sound.name.includes(phrase) || (sound.tags || []).some((tag) => tag.includes(phrase));
+  const { name, tags } = sound;
+  const tagsAndNameJoinedString = [name, ...tags ?? []].join('').toLowerCase();
```
DzikowskiW, line 9: I would use `join(' ')` with a space inside, otherwise we may get a match for a phrase that consists of last letters of name concatenated with first letters of a tag.

packages/fm-player/src/components/SoundBoard.tsx
```diff
@@ -5,7 +5,9 @@ import SoundButton from './SoundButton';
...
+  return phrase.toLowerCase().split(' ').every((keyword) => tagsAndNameJoinedString.includes(keyword));
```
DzikowskiW, line 10: Just an observation: I was trying to come up with an idea that would not include concatenating strings, but `.includes()` right here makes it challening to replace it with a proper data structure that would be easily readable. Let's stick to what you came up with :).

DzikowskiW (changes requested):
Już pisałem propozycję, żeby wykorzystać Sety, ale musielibyśmy mieć

packages/fm-player/src/components/SoundBoard.tsx
gorzalczany: Nice catch. I used "" to shorten the search string but did not thought about the case you mentioned.

Deseteral (approved):

PR #184, opened 2022-10-23
Adding tags in xSoundUploadForm; Login bg fix

mswiechowicz:
Now we can add tags just in upload form. Also fixed login bg unresolved path

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,8 +200,8 @@ Adds new sound file to X Sounds database.
 
 **Request** \
 Requires content type to be `multipart/form-data` with fields:
-- `soundName`: name of the sound to be added (like _"bruh"_
-- `tags` (optional): tag list separated with comma (like _"jp2, jan pawel, jean paul"_)
+- `soundName`: name of the sound to be added (like _"cool sound"_)
+- `tags` (optional): comma separated list of tags (like _"tag1, tag2, tag3"_)
 - `soundFile`: sound [File](https://developer.mozilla.org/en-US/docs/Web/API/File) ideally in mp3 or wav format
 
 **Response**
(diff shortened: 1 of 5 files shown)


packages/service/src/api/x-sounds/x-sounds-controller.ts
```diff
@@ -26,19 +26,22 @@ class XSoundsController {
   async addXSound(
     @UploadedFile('soundFile') soundFile: any,
     @BodyParam('soundName') soundName: string,
+    @BodyParam('tags') tags: string,
```
Deseteral, line 29: - IMO z punktu widzenia API to pole powinno być opcjonalne. 
- Dodaj do readme że ten endpoint ma nowe pole

README.md
```diff
@@ -200,7 +200,8 @@ Adds new sound file to X Sounds database.
 
 **Request** \
 Requires content type to be `multipart/form-data` with fields:
-- `soundName`: name of the sound to be added (like _"bruh"_)
+- `soundName`: name of the sound to be added (like _"bruh"_
+- `tags` (optional): tag list separated with comma (like _"jp2, jan pawel, jean paul"_)
```
Deseteral, line 204: 1. Let's keep the readme context free
2. "comma separated list" would sound better

So I suggest something like this:

```suggestion
- `tags` (optional): comma separated list of tags (like _"tag1,tag2,tag3"_)
```

Same thing in the line above - instead of `bruh`, more generic name would be better like `cool sound` or something. You're also missing closing bracket on that line.

Deseteral (changes requested):

README.md
mswiechowicz: Fixed

Deseteral (approved):

Issue #198, opened 2023-07-21
Sound preview on soundboard  


gorzalczany:
Add hover action over a sound pill to display a button to play  sound locally.

Deseteral:
Closed by #207

PR #207, opened 2023-10-14
XSound will play locally while holding down cmd or alt key

mswiechowicz:


packages/client/src/sound-board/service/soundboard-service.ts
```diff
@@ -1,3 +1,5 @@
+const audioClient = new Audio();
```
mswiechowicz, line 1: It's here because it would be nice to have just one instance of the new Audio

mswiechowicz:
resolves #198

packages/client/src/sound-board/service/soundboard-service.ts
```diff
@@ -1,5 +1,6 @@
+const AudioClient = new Audio();
```
mswiechowicz, line 1: It's here because it would be nice to have just one instance of the new Audio

Deseteral: This is fine :) But I think it should be lower case

```suggestion
const audioClient = new Audio();
```

packages/client/src/sound-board/service/soundboard-service.ts
```diff
@@ -9,6 +10,9 @@ async function queueXSound(name: string) {
   });
 }
 
-export {
-  queueXSound,
-};
+function playXSoundLocally(url: string) {
+  AudioClient.src = url;
+  AudioClient.play();
```
Deseteral, line 15: Shouldn't this technically be called inside a `load` event?
I think that there is a possibility for `.play()` to be called before the sound is downloaded.

Deseteral (approved):

PR #210, opened 2023-10-18
Added soundboard layer widget on fm player view

kstodolak:


package.json
```diff
@@ -25,4 +25,4 @@
     "@nrwl/workspace": "14.8.5",
     "nx": "14.8.5"
   }
-}
+}
```
Deseteral, line 28: This is not the way

Deseteral (changes requested):

package.json
kstodolak: 💩

packages/client/src/fm-player/components/SoundBoardWidget.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { SoundIcon } from '../icons/SoundIcon';
+import { SoundBoard } from '../../sound-board/components/SoundBoard';
+
+function SoundBoardWidget() {
+  const soundBoardLayer = React.useRef<HTMLDivElement>(null);
+
+  const toggleSoundBoardView = () => {
+    if (soundBoardLayer.current) {
+      soundBoardLayer.current.classList.toggle('expanded');
```
Deseteral, line 10: This is not the way

Deseteral (changes requested):

Deseteral (approved):

PR #219, opened 2023-11-04
Dockerized

Deseteral:


fly.toml
```diff
@@ -11,15 +6,9 @@ kill_timeout = "5s"
 [experimental]
   auto_rollback = true
 
-[build]
```
gorzalczany, line 14: So now we are leveraging default behavior of [looking for dockerfile in application root](https://fly.io/docs/reference/configuration/#specify-a-dockerfile)? Don't we want to describe it explicitely not to leave space for guesses?

Deseteral: Sure, we could explicitly specify it. I don't have a preference here.

gorzalczany: Also fly's gotchas:
> When specifying a local Dockerfile, make sure it’s not excluded from the Docker build context in your .dockerignore.

Deseteral: Done

gorzalczany (approved):

gorzalczany (approved):

PR #222, opened 2023-11-20
User updates

DzikowskiW:
Added Users counter in Player view:
image

### DIFF
--- a/packages/client/src/fm-player/services/connected-users-service.ts
+++ b/packages/client/src/fm-player/services/connected-users-service.ts
@@ -10,7 +10,7 @@ function getUsers(): string[] {
 
 function setUsers(users: string[]): void {
   connectedUsers = users;
-  emitter.emit('change', connectedUsers);
+  emitter.emit('change', getUsers());
 }
 
 function onUsersChange(callback: Handler<string[]>) {
--- a/packages/client/src/fm-player/services/connected-users-service.ts
+++ b/packages/client/src/fm-player/services/connected-users-service.ts
@@ -1,0 +1,24 @@
+import mitt, { Emitter, Handler } from 'mitt';
+
+const emitter: Emitter = mitt();
+
+let connectedUsers: string[] = [];
+
+function getUsers(): string[] {
+  return connectedUsers;
+}
+
+function setUsers(users: string[]): void {
+  connectedUsers = users;
+  emitter.emit('change', connectedUsers);
+}
+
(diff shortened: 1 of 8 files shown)


packages/client/src/fm-player/services/connected-users-service.ts
```diff
@@ -0,0 +1,24 @@
+import mitt, { Emitter, Handler } from 'mitt';
+
+const emitter: Emitter = mitt();
+
+let connectedUsers: string[] = [];
+
+function getUsers(): string[] {
+  return connectedUsers;
```
Deseteral, line 8: I feel like this function should return a copy of that array, so that other parts of the application would not be able to mutate it's global state. 

Essentially make this array be only internally mutable from this file.

Same thing down below in the event emitter.

Deseteral: But I do admit that this is my inner ~demons~ Rust-isms showing up.

Deseteral (approved):

packages/client/src/fm-player/services/connected-users-service.ts
DzikowskiW: I agree, I added a fix

packages/client/src/fm-player/services/connected-users-service.ts
```diff
@@ -0,0 +1,24 @@
...
+
+let connectedUsers: string[] = [];
+
+function getUsers(): string[] {
+  return [...connectedUsers];
+}
+
+function setUsers(users: string[]): void {
+  connectedUsers = users;
+  emitter.emit('change', connectedUsers);
```
Deseteral, line 13: Same thing here
```suggestion
  emitter.emit('change', [...connectedUsers]);
```

packages/client/src/fm-player/services/connected-users-service.ts
```diff
@@ -0,0 +1,24 @@
+import mitt, { Emitter, Handler } from 'mitt';
+
+const emitter: Emitter = mitt();
+
+let connectedUsers: string[] = [];
+
+function getUsers(): string[] {
+  return [...connectedUsers];
```
Deseteral, line 8: We actually could do some TypeScript magic here:
```suggestion
  return [...connectedUsers] as const;
```

It marks the array as mutable, this way type checker will scream if we every try to change it.

It also changes the return type of that function to `readonly string[]`.

Deseteral (approved):

PR #224, opened 2023-11-22
Current users list directly in a player

DzikowskiW:
image

### DIFF
--- a/packages/client/src/fm-player/components/UsersButton.tsx
+++ b/packages/client/src/fm-player/components/UsersButton.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { UsersIcon } from '../icons/UsersIcon';
 import * as ConnectedUsersService from '../services/connected-users-service';
-import { CloseIconSolid } from '../icons/CloseIconSolid';
 
 function UsersButton() {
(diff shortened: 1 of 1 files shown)


packages/client/src/fm-player/components/UsersButton.tsx
```diff
@@ -13,17 +14,31 @@ function UsersButton() {
...
+        className="absolute bg-green-400 rounded-xl p-3 hover:bg-green-600 text-2xl font-bold leading-none"
+      >
+        <div className="flex">
+          <UsersIcon />
+          <span className="pl-1">{currentUsers.length}</span>
+        </div>
+        {isExtended && (
+        <ul className="text-base font-normal leading-normal pt-1">
+          {currentUsers.map((user, index) => (
+            // eslint-disable-next-line react/no-array-index-key
```
DzikowskiW, line 34: we don't have a better key than index at this point, therefore I disabled this check for this case

Deseteral: We do have a better a key - the user name itself :D They're unique, that should do.

Deseteral (approved):

PR #228, opened 2023-11-29
No one should suffer from Tailwind

Deseteral:


Deseteral:
We're sunsetting current frontend. This PR is no longer necessary.

PR #229, opened 2023-12-08
split play-pause functionality into separate commands

gorzalczany:
#Why:
- avoid confusion when some players changed state locally (local pause)
- enable "force play" when some players have problem
- enable play again on finished song that was last in the queue

### DIFF
--- a/packages/client/src/fm-player/services/event-stream-client.ts
+++ b/packages/client/src/fm-player/services/event-stream-client.ts
@@ -39,8 +39,12 @@ function connect(): (() => void) {
         SpeechService.say(eventData.text);
         break;
 
-      case 'PlayPauseEvent':
-        YouTubePlayerService.playPause();
+      case 'PauseEvent':
+        YouTubePlayerService.pause();
+        break;
+
+      case 'ResumeEvent':
+        YouTubePlayerService.resume();
         break;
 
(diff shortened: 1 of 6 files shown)


packages/service/src/domain/commands/processors/playback-play-pause-command.ts
```diff
@@ -2,23 +2,14 @@ import { Command } from '@service/domain/commands/model/command';
 import { CommandProcessingResponse, CommandProcessingResponseBuilder } from '@service/domain/commands/model/command-processing-response';
 import { CommandParameters, CommandParametersBuilder, CommandProcessor } from '@service/domain/commands/model/command-processor';
 import { RegisterCommand } from '@service/domain/commands/registry/register-command';
-import { PlayPauseEvent } from '@service/event-stream/model/events';
-import { PlayerEventStream } from '@service/event-stream/player-event-stream';
 import { Service } from 'typedi';
 
 @RegisterCommand
 @Service()
 class PlayPauseCommand extends CommandProcessor {
```
Deseteral, line 9: Why not just remove it and be done with it?

Deseteral: Fine by me :)

PR #245, opened 2024-10-16
xsounds filters added

mswiechowicz:


packages/client-beta/src/apps/Soundboard/Soundboard.tsx
```diff
@@ -25,9 +28,34 @@ function Soundboard() {
   createEffect(() => {
     fetch('/api/x-sounds')
       .then((res) => res.json())
-      .then((res) => setXsounds(res.sounds));
+      .then((res) => {
+        setXsounds(res.sounds);
+        setFilteredXSounds(res.sounds);
+      });
+
+    fetch('/api/x-sounds/tags')
```
Deseteral, line 36: I feel like fetching should be done in some kind of client, so that in components we don't call `fetch` and don't have to worry about JSON parsing - just calling a method like `getXSoundTags()` that returns a Promise with a list of tags.

packages/client-beta/src/services/sounds-filtering.ts
```diff
@@ -0,0 +1,24 @@
...
+    const isGoodMatchB = b.name.startsWith(phrase);
+
+    if (isGoodMatchA !== isGoodMatchB) {
+      return isGoodMatchA ? -1 : 1;
+    }
+
+    return a.name.localeCompare(b.name, undefined, { sensitivity: 'base' }) || b.timesPlayed - a.timesPlayed;
+  };
+}
+
```
Deseteral, line 24: I'm not saying anything, I'm not suggesting anything - those methods looks like a great opportunity to write some tests 😉

Deseteral (approved):

packages/client-beta/src/apps/Soundboard/Soundboard.tsx
```diff
@@ -25,9 +28,34 @@ function Soundboard() {
...
+      .then((res) => setTags(res.tags));
   });
 
+
+  const onSearchChange = (e: Event) => {
+    const phrase = (e.target as HTMLInputElement).value;
+    const name = (e.target as HTMLInputElement).name;
+
+    if(name === 'phrase') {
+      (document.getElementsByName('tags')[0] as HTMLInputElement).value = '';
```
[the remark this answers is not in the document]
mswiechowicz: It's because getElementsByName returns Element type, and Element type doesnt have a "value".
image
Maybe a better way to catch this would be to give these inputs an id. Then we don't have to do [0]. Or maybe querySelector('input[name=tags'). Didn't think about it :D I'd still have to do that casting, because each returns an Element

packages/client-beta/src/services/sounds-filtering.ts
mswiechowicz: Thanks to @gorzalczany for these beautifully filtering features :D I copied them from a previous soundboard

packages/client-beta/src/apps/Soundboard/Soundboard.tsx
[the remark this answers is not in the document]
Deseteral: I don't have a problem with all of that, I was just surprised that there is no way to get the native HTML element using Solid and we have to use document querying.

Like React has `ref={this.myElement}` or something like that and Solid does not?

PR #249, opened 2024-10-25
beta / feature / dropbox file upload

gorzalczany:
### What:
- add dropbox file upload logic

### DIFF
--- a/service/build.gradle.kts
+++ b/service/build.gradle.kts
@@ -29,6 +29,7 @@ dependencies {
     implementation("io.ktor:ktor-client-okhttp")
     implementation("io.ktor:ktor-client-auth")
     implementation("io.ktor:ktor-client-content-negotiation")
+
     implementation("io.ktor:ktor-server-core-jvm")
     implementation("io.ktor:ktor-server-resources-jvm")
     implementation("io.ktor:ktor-server-host-common-jvm")
--- a/service/build.gradle.kts
+++ b/service/build.gradle.kts
@@ -27,6 +27,7 @@ repositories { mavenCentral() }
 dependencies {
     implementation("io.ktor:ktor-client-core")
     implementation("io.ktor:ktor-client-cio")
+    implementation("io.ktor:ktor-client-auth")
     implementation("io.ktor:ktor-client-content-negotiation")
     implementation("io.ktor:ktor-server-core-jvm")
     implementation("io.ktor:ktor-server-resources-jvm")
--- a/service/build.gradle.kts
+++ b/service/build.gradle.kts
@@ -25,6 +25,9 @@ application {
 repositories { mavenCentral() }
 
 dependencies {
+    implementation("io.ktor:ktor-client-core")
+    implementation("io.ktor:ktor-client-cio")
+    implementation("io.ktor:ktor-client-content-negotiation")
     implementation("io.ktor:ktor-server-core-jvm")
     implementation("io.ktor:ktor-server-resources-jvm")
     implementation("io.ktor:ktor-server-host-common-jvm")
--- a/service/build.gradle.kts
+++ b/service/build.gradle.kts
@@ -26,7 +26,7 @@ repositories { mavenCentral() }
 
 dependencies {
     implementation("io.ktor:ktor-client-core")
-    implementation("io.ktor:ktor-client-cio")
+    implementation("io.ktor:ktor-client-okhttp")
     implementation("io.ktor:ktor-client-auth")
     implementation("io.ktor:ktor-client-content-negotiation")
     implementation("io.ktor:ktor-server-core-jvm")
--- a/service/src/main/kotlin/xyz/lebkuchenfm/api/xsounds/XSoundsEndpoint.kt
+++ b/service/src/main/kotlin/xyz/lebkuchenfm/api/xsounds/XSoundsEndpoint.kt
@@ -17,6 +17,8 @@ import xyz.lebkuchenfm.domain.xsounds.XSound
 import xyz.lebkuchenfm.domain.xsounds.XSoundsService
 import xyz.lebkuchenfm.external.storage.FileStorage
 import xyz.lebkuchenfm.external.storage.Storage
+import java.io.File
+import java.util.UUID
 
 fun Route.xSoundsRouting(xSoundsService: XSoundsService) {
     route("/x-sounds") {
@@ -25,6 +27,7 @@ fun Route.xSoundsRouting(xSoundsService: XSoundsService) {
             val response = XSoundsResponse(sounds.map { it.toResponse() })
             call.respond(response)
         }
+
         get("/tags") {
 
         }
@@ -32,7 +35,7 @@ fun Route.xSoundsRouting(xSoundsService: XSoundsService) {
         post {
             var soundName = ""
             var tagsString = ""
-            var fileName = ""
+            var fileName = UUID.randomUUID().toString()
             val multipartData = call.receiveMultipart()
             var fileBytes: ByteArray? = null
 
@@ -40,14 +43,16 @@ fun Route.xSoundsRouting(xSoundsService: XSoundsService) {
                 when (part) {
                     is PartData.FormItem -> {
                         if (part.name == "soundName") {
+                            // TODO: check if soundName provided and valid
                             soundName = part.value
                         } else if (part.name == "tags") {
+                            // TODO: parse tags
                             tagsString = part.value
                         }
                     }
 
                     is PartData.FileItem -> {
-                        fileName = part.originalFileName as String
+                        part.originalFileName?.let { fileName = it }
                         fileBytes = part.provider().readRemaining().readByteArray()
                     }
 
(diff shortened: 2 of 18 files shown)


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxFileStorage.kt
```diff
@@ -0,0 +1,132 @@
...
+        val sharedFile: ShareFileResponse =
+            client.post("https://api.dropboxapi.com/2/sharing/create_shared_link_with_settings") {
+                setBody(ShareFile(uploadFile.pathDisplay, ShareFileSettings("viewer", "public", true)))
+                contentType(ContentType.Application.Json)
+            }.body()
+
+        val resourceUrl =
+            URLBuilder(sharedFile.url).apply {
+                this.host = "dl.dropboxusercontent.com"
+                this.parameters.remove("dl")
```
Deseteral, line 71: You can remove `this.` inside `apply`. I think it's clearer, but I will leave it up to you.

service/build.gradle.kts
```diff
@@ -25,6 +25,10 @@ application {
 repositories { mavenCentral() }
 
 dependencies {
+    implementation("io.ktor:ktor-client-core")
+    implementation("io.ktor:ktor-client-okhttp")
+    implementation("io.ktor:ktor-client-auth")
+    implementation("io.ktor:ktor-client-content-negotiation")
```
Deseteral, line 31: Leave empty line below this to separate ktor-client from ktor-server.

service/src/main/kotlin/xyz/lebkuchenfm/domain/xsounds/XSound.kt
```diff
@@ -3,6 +3,8 @@ package xyz.lebkuchenfm.domain.xsounds
 typealias XSoundId = String
```
Deseteral, line 3: If you have removed XSound.id then remove unused alias here as well.

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxFileStorage.kt
```diff
@@ -0,0 +1,132 @@
...
+
+class DropboxFileStorage(config: ApplicationConfig) {
+    private val client = prepareClient()
+    private val bearerTokenStorage = mutableListOf<BearerTokens>()
+    private val refreshToken = config.property(DROPBOX_REFRESH_TOKEN_PROPERTY_PATH).getString()
+    private val appKey = config.property(DROPBOX_APP_KEY_PROPERTY_PATH).getString()
+    private val appSecret = config.property(DROPBOX_APP_SECRET_PROPERTY_PATH).getString()
+
+    /**
+     * [path] should contain destination folder, file name and its extension
```
Deseteral, line 37:
```suggestion
     * [path] must contain destination folder, file name and its extension
```

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/models/ShareFile.kt
```diff
@@ -0,0 +1,22 @@
+package xyz.lebkuchenfm.external.storage.dropbox.models
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class ShareFileSettings(
```
Deseteral, line 7: I would suggest to rename all of the data classes that we use to communicate with Dropbox with a `DropboxClassName` prefix.

```suggestion
data class DropboxShareFileSettings(
```

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/repos/XSoundsMongoRepository.kt
```diff
@@ -14,13 +14,30 @@ class XSoundsMongoRepository(database: MongoDatabase) : XSoundsRepository {
...
     }
+
+    override suspend fun insert(sound: XSound) {
+        collection.insertOne(XSoundEntity(sound))
+    }
 }
 
 data class XSoundEntity(
-    @BsonId val id: ObjectId,
+    @BsonId val id: ObjectId? = null,
```
Deseteral, line 24: Wait, why null? ObjectId cannot be null, all entires in our database have `_id` field 🤔

service/src/main/resources/application.yaml
```diff
@@ -8,3 +8,10 @@ ktor:
 storage:
   mongodb:
     connectionString: "$MONGODB_URL:mongodb://localhost:27017"
+  dropbox:
+    auth:
+      refreshToken: "$DROPBOX_REFRESH_TOKEN"
+      appKey: "$DROPBOX_APP_KEY"
+      appSecret: "$DROPBOX_APP_SECRET"
+    paths:
+      xSounds: "$DROPBOX_X_SOUNDS_PATH:/lebkuchenFM/x-sounds/"
```
Deseteral, line 17: Actually the path should be just `/xsounds/` since we use the Dropbox Apps integration it will be automatically put in the right LebkuchenFM name space.

Deseteral: We don't need to get this from env, let's just leave it as 
```suggestion
      xSounds: "/xsounds/"
```

Deseteral (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/api/xsounds/XSoundsEndpoint.kt
```diff
@@ -15,6 +22,43 @@ fun Route.xSoundsRouting(xSoundsService: XSoundsService) {
...
+            var tags: List<String> = emptyList()
+            var fileBytes: ByteArray? = null
+
+            val multipartData = call.receiveMultipart()
+
+            multipartData.forEachPart { part ->
+                when (part) {
+                    is PartData.FormItem -> {
+                        if (part.name == "soundName") {
+                            // TODO: check if provided soundName is valid
```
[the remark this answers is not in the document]
gorzalczany: eg. null, empty, special characters, no file extension in name
so i propose
```
// TODO: `soundName` validation ((null, empty, special characters, no file extension in name))
```

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxFileStorage.kt
```diff
@@ -0,0 +1,132 @@
...
+                append("grant_type", "refresh_token")
+            },
+        )
+
+    private suspend fun getInitialTokens() {
+        val tokenInfo: TokenInfo =
+            HttpClient(OkHttp) {
+                install(ContentNegotiation) { json() }
+            }.post("https://api.dropbox.com/oauth2/token") {
+                setBody(getRefreshTokenRequestBody())
```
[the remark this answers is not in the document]
gorzalczany: But it is reused in `refreshTokens` method in Auth plugin

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/models/ShareFile.kt
gorzalczany: those are already in dropbox package but ok

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/repos/XSoundsMongoRepository.kt
gorzalczany: Because we have to insert entity to database first and only after that we get id.

service/src/main/resources/application.yaml
gorzalczany: are you sure about this namespace? Does not work like that for me.
Thats why I've added possibility to set whatever path from env.

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxFileStorage.kt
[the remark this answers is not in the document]
Deseteral: Oh okay, I didn't see it ☺️

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/repos/XSoundsMongoRepository.kt
Deseteral: Okay, I might be interesting to follow up later 👍

service/src/main/resources/application.yaml
Deseteral: That's how we did it until now. If it does not work for you then leave the env part, but please change the default value.

service/src/main/kotlin/xyz/lebkuchenfm/api/xsounds/XSoundsEndpoint.kt
```diff
@@ -15,6 +23,42 @@ fun Route.xSoundsRouting(xSoundsService: XSoundsService) {
...
+
+                    else -> {}
+                }
+                part.dispose()
+            }
+
+            fileBytes?.let { bytes ->
+                // TODO: pass authenticated user name as "addedBy"
+                val sound = xSoundsService.addNewXSound(soundName, tags, bytes)
+                val response = Json.encodeToString(XSoundResponse.serializer(), sound.toResponse())
```
Deseteral, line 58: You don't have to serialize manually. You can just pass the object to `call.respond` as said in the [docs](https://ktor.io/docs/server-responses.html#object).

Deseteral (changes requested):

Deseteral (approved):

PR #252, opened 2024-10-26
[beta backend] add option to get all xsounds tags

gorzalczany:


service/src/main/kotlin/xyz/lebkuchenfm/domain/xsounds/XSoundsService.kt
```diff
@@ -19,4 +19,8 @@ class XSoundsService(private val repository: XSoundsRepository, private val file
         repository.insert(readySound)
         return readySound
     }
+
+    suspend fun getAllUniqueTags(): List<String> {
```
Deseteral, line 23:
```suggestion
    suspend fun findAllUniqueTags(): List<String> {
```
because of [this](https://tuhrig.de/find-vs-get/).

Deseteral (changes requested):

Deseteral (approved):

PR #254, opened 2024-10-26
[beta backend] add songs endpoint

gorzalczany:


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/repos/XSoundsMongoRepository.kt
```diff
@@ -18,7 +18,7 @@ import xyz.lebkuchenfm.domain.xsounds.XSoundsRepository
 
 class XSoundsMongoRepository(database: MongoDatabase) : XSoundsRepository {
     private val collection = database.getCollection<XSoundEntity>("x")
-    private val sortByName = Sorts.ascending(XSound::name.name)
+    private val sortByName = Sorts.ascending(XSoundEntity::name.name)
```
Deseteral, line 21: Good catch

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/repos/SongsMongoRepository.kt
```diff
@@ -0,0 +1,50 @@
...
+
+data class SongEntity(
+    @BsonId val id: ObjectId? = null,
+    val name: String,
+    val youtubeId: String,
+    val trimStartSeconds: Int?,
+    val trimEndSeconds: Int?,
+    val timesPlayed: Int,
+) {
+    constructor(song: Song) : this(
```
Deseteral, line 32: Why do we need that?

Deseteral (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/repos/SongsMongoRepository.kt
gorzalczany: we don't. At least yet 👍🏻

Deseteral (approved):

PR #259, opened 2024-10-30
[beta backend] improve error handling in youtube client

gorzalczany:
What:
- use result type
- fix yaml configuration optional value
- use logger

and the change that I'm not sure if I like:
- use `DefaultRequest` ktor pluggin instead of custom one


service/src/main/resources/application.yaml
```diff
@@ -17,6 +17,6 @@ storage:
       xSounds: "$DROPBOX_X_SOUNDS_PATH:/xsounds/"
 
 youtube:
-  apiKey: "$YOUTUBE_API_KEY"
+  apiKey: $?YOUTUBE_API_KEY
```
Deseteral, line 20: 🤲

Deseteral: Shame that it's not documented anywhere :(

service/src/main/kotlin/xyz/lebkuchenfm/external/youtube/YoutubeClient.kt
```diff
@@ -1,44 +1,80 @@
...
+
 class YoutubeClient(config: ApplicationConfig) {
     private val youtubeClient: HttpClient by lazy { prepareHttpClient() }
     private val apiKey: String? by lazy { config.propertyOrNull("youtube.apiKey")?.getString() }
 
-    suspend fun getVideoName(id: String): String {
-        val response: YoutubeVideosResponse = youtubeClient.get("/videos") {
+    suspend fun getVideoName(id: String): Result<String, YoutubeClientError> {
+        if (apiKey == null) {
+            logger.warn { "Missing Youtube API key" }
```
Deseteral, line 32:
```suggestion
            logger.warn { "Missing Youtube API key." }
```

service/src/main/kotlin/xyz/lebkuchenfm/external/youtube/YoutubeClient.kt
```diff
@@ -1,44 +1,80 @@
...
             parameters {
                 parameter("id", id)
                 parameter("part", "snippet")
                 parameter("filer", "id")
             }
-        }.body()
+        }
+
+        if (response.status != HttpStatusCode.OK) {
+            logger.error { "YoutubeClient failed with status: ${response.status} \n ${response.request.url}" }
```
Deseteral, line 45:
```suggestion
            logger.error { "YoutubeClient failed with status: ${response.status} \n ${response.request.url}." }
```

service/src/main/kotlin/xyz/lebkuchenfm/external/youtube/YoutubeClient.kt
```diff
@@ -1,44 +1,80 @@
...
-            install(YoutubeHttpClientPlugin) {
-                apiKey = this@YoutubeClient.apiKey
-            }
             install(ContentNegotiation) { json(Json { ignoreUnknownKeys = true }) }
+            install(DefaultRequest) {
+                url {
+                    protocol = URLProtocol.HTTPS
+                    host = "www.googleapis.com"
+                    path("/youtube/v3/")
+                    this@YoutubeClient.apiKey?.let { parameters.append("key", it) }
```
Deseteral, line 65: 🤔 I'm thinking aloud here.
If there is no key, we shouldn't be able to create a client. Isn't this a illegal state of application? If the key is not present, then we should not create a client?

Deseteral (approved):

Deseteral (approved):

PR #261, opened 2024-11-03
[beta backend] add discord client

gorzalczany:
### What:
- add discord client with `kord` as dependency

### DIFF
--- a/service/build.gradle.kts
+++ b/service/build.gradle.kts
@@ -47,6 +47,8 @@ dependencies {
 
     implementation("org.mongodb:mongodb-driver-kotlin-coroutine:$mongo_driver_version")
 
+    implementation("dev.kord:kord-core:0.15.0")
+
     testImplementation("io.ktor:ktor-server-test-host-jvm")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit:$kotlin_version")
 }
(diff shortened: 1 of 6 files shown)


service/build.gradle.kts
```diff
@@ -47,6 +47,8 @@ dependencies {
 
     implementation("org.mongodb:mongodb-driver-kotlin-coroutine:$mongo_driver_version")
 
+    implementation("dev.kord:kord-core:0.15.0")
```
Deseteral, line 50: Please extract the version to `gradle.properties`.

service/src/main/resources/logback.xml
```diff
@@ -11,4 +11,5 @@
     <logger name="io.netty" level="INFO"/>
     <logger name="org.mongodb.driver" level="INFO" />
     <logger name="io.ktor" level="DEBUG" />
+    <logger name="dev.kord" level="TRACE" />
```
Deseteral, line 14: I don't think this does anything. Trace is the root level.

Deseteral (changes requested):

service/src/main/resources/logback.xml
gorzalczany: As you said. But I thought we might like to see it explicitly for libs.

Deseteral (approved):

PR #262, opened 2024-11-03
Initial auth

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/Application.kt
```diff
@@ -56,20 +70,63 @@ fun Application.module() {
...
     val discordClient = DiscordClient(environment.config, commandExecutorService)
-
     launch { discordClient.start() }
 
     install(ContentNegotiation) {
         json()
     }
 
+    install(Sessions) {
+        val days30 = 30 * 24 * 60 * 60 * 1000L
```
gorzalczany, line 80: thats in millisecond but variable below is `maxAgeInSeconds`

service/src/main/kotlin/xyz/lebkuchenfm/api/auth/AuthEndpoint.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.ktor.server.sessions.get
+import io.ktor.server.sessions.sessions
+import io.ktor.server.sessions.set
+import kotlinx.serialization.Serializable
+import xyz.lebkuchenfm.domain.auth.UserSession
+
+fun Route.authRouting() {
+    route("/auth") {
+        get {
+            val session = call.sessions.get<UserSession>()!!
```
gorzalczany, line 21: force?

gorzalczany (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/api/auth/AuthEndpoint.kt
Deseteral: Yes. Technically it should never be null. So at least while we're in development mode I would like it to throw Null Pointer Exception so we can debug the case when it is.

service/src/main/kotlin/xyz/lebkuchenfm/Application.kt
Deseteral: You're absolutely right!

gorzalczany (approved):

PR #263, opened 2024-11-03
Forward compatibility of login form

Deseteral:
Instead of JSON based login request we can move to form-data based login flow. This will allow us to test the upcoming v3 version of backend service with current client.

### DIFF
--- a/packages/client/src/services/user-account-service.ts
+++ b/packages/client/src/services/user-account-service.ts
@@ -1,14 +1,13 @@
-import { AuthRequestDto } from 'lebkuchen-fm-service';
 import { redirectTo } from './redirect-to';
 
 export function userLogin(username: string, password: string) {
-  const data: AuthRequestDto = { username, password };
+  const body = new URLSearchParams();
+  body.set('username', username);
+  body.set('password', password);
+
   const options: RequestInit = {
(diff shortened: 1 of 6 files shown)


packages/client/src/services/user-account-service.ts
```diff
@@ -1,14 +1,13 @@
-import { AuthRequestDto } from 'lebkuchen-fm-service';
 import { redirectTo } from './redirect-to';
 
 export function userLogin(username: string, password: string) {
-  const data: AuthRequestDto = { username, password };
+  const body = new URLSearchParams();
```
gorzalczany, line 4: shouldnt we use `FormData` instead of `URLSearchParams`?

Deseteral: Actually - no. While the new backend accepts both `application/x-www-form-urlencoded` and `multipart/form-data` the old backend (after this change) can only support `application/x-www-form-urlencoded` (without much work). 
Besides - for simple forms like this one we [should be using `application/x-www-form-urlencoded` anyway](https://stackoverflow.com/questions/4007969/application-x-www-form-urlencoded-or-multipart-form-data).

gorzalczany (approved):

PR #265, opened 2024-11-04
fm x implementation for the new service

DzikowskiW:


service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/XCommandProcessor.kt
```diff
@@ -27,7 +28,7 @@ class XCommandProcessor(private val xSoundsService: XSoundsService, private val
         val soundName = command.rawArgs
             ?: return error("You have to provide sound name.", logger)
 
-        val xSound = xSoundsService.getByName(soundName)
+        val xSound = runBlocking { xSoundsService.getByName(soundName) }
```
gorzalczany, line 31: maybe `fun execute` should `suspend` 🤔

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/XSoundsMongoRepository.kt
```diff
@@ -51,6 +52,10 @@ class XSoundsMongoRepository(database: MongoDatabase) : XSoundsRepository {
         ).first()
         return result.tagsSet
     }
+
+    override suspend fun findByName(name: String): XSound? {
+        return collection.find(eq(XSound::name.name, name)).firstOrNull()?.toDomain()
```
gorzalczany, line 57: `XSoundEntity::name.name, name` - we should use mongo's entity here
Please fix also filter in line 32 🙏🏻

gorzalczany (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/XSoundsMongoRepository.kt
```diff
@@ -39,6 +40,10 @@ class XSoundsMongoRepository(database: MongoDatabase) : XSoundsRepository {
         collection.insertOne(XSoundEntity(sound))
     }
 
+    override suspend fun replace(sound: XSound) {
+        collection.replaceOne(eq(XSoundEntity::name.name, sound.name), XSoundEntity(sound))
```
gorzalczany, line 44: WDYT about incrementing count when finding? 

```
override suspend fun findByName(name: String, updateCount: Boolean): XSound? {
    val filter = eq(XSoundEntity::name.name, name)

    return if (updateCount) {
        val increment = inc(XSoundEntity::timesPlayed.name, 1)
        val options = FindOneAndUpdateOptions().returnDocument(ReturnDocument.BEFORE)
        collection.findOneAndUpdate(filter, increment, options)?.toDomain()
    } else {
        collection.find(filter).firstOrNull()?.toDomain()
    }
}
```

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/XSoundsMongoRepository.kt
```diff
@@ -65,12 +74,16 @@ data class XSoundEntity(
         null,
         sound.name,
         sound.url,
-        0,
+        sound.timesPlayed,
```
gorzalczany, line 77: I maintain the opinion that when we create new entity then timesPlayed should be `0` ;p

Deseteral: First - I don't like that we have this as a constructor. We should have a `fromDomain` method (or something like that).
Second - @gorzalczany I don't agree. New entity should have `timesPlayed=0`, but it's not the responsibility of the repository to decide if the sound is new or not. That's done in the service. Here you don't even know that the sound is new. It might as well be imported from somewhere, or something like that. Service passes a domain object, repository saves it, does not think about where it comes from, or what it is exactly.

Deseteral: This constructor/method should only be doing mapping from domain object to database entity.

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/XSoundsMongoRepository.kt
Deseteral: As per discussion on Discord - we should be doing `findOneAndUpdate` (because we live in multithreaded environment) but not in `findByName` only in different dedicated method (`incrementPlayCountByName`?)

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/XSoundsMongoRepository.kt
```diff
@@ -60,6 +57,13 @@ class XSoundsMongoRepository(database: MongoDatabase) : XSoundsRepository {
     override suspend fun findByName(name: String): XSound? {
         return collection.find(eq(XSoundEntity::name.name, name)).firstOrNull()?.toDomain()
     }
+
+    override suspend fun incrementPlayCount(sound: XSound) {
```
Deseteral, line 61: Let's return the updated `XSound?` here (just convert what `findOneAndUpdate` returns to domain object).

service/src/main/kotlin/xyz/lebkuchenfm/domain/xsounds/XSoundsService.kt
```diff
@@ -24,10 +24,10 @@ class XSoundsService(private val repository: XSoundsRepository, private val file
         return repository.findByName(soundName)
     }
 
-    suspend fun incrementPlayCount(soundName: String) {
+    suspend fun markAsPlayed(soundName: String) {
         val xSound = this.getByName(soundName)
```
Deseteral, line 28: We don't have to do that. Let repository update the entity if it exists.

service/src/main/kotlin/xyz/lebkuchenfm/domain/xsounds/XSoundsService.kt
```diff
@@ -24,10 +24,10 @@ class XSoundsService(private val repository: XSoundsRepository, private val file
...
         xSound?.let {
-            repository.replace(it.copy(timesPlayed = it.timesPlayed + 1))
+            repository.incrementPlayCount(it)
```
Deseteral, line 30: Check if the `incrementPlayCount` returns null. If it does we would like to log that we couldn't increment the play count for some reason.

service/src/main/kotlin/xyz/lebkuchenfm/domain/xsounds/XSoundsService.kt
```diff
@@ -25,9 +25,6 @@ class XSoundsService(private val repository: XSoundsRepository, private val file
     }
 
     suspend fun markAsPlayed(soundName: String) {
-        val xSound = this.getByName(soundName)
-        xSound?.let {
-            repository.incrementPlayCount(it)
-        }
+        repository.incrementPlayCount(soundName)
```
Deseteral, line 28: Also please add log that something went wrong:
```suggestion
        repository.incrementPlayCount(soundName)?.also { 
            logger.error { "Could not increment play count for sound $soundName" }
        }
```

Deseteral (approved):

gorzalczany (approved):

PR #267, opened 2024-11-06
add-tag - added an empty command processor

DzikowskiW:


service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/TagAddCommandProcessor.kt
```diff
@@ -0,0 +1,35 @@
...
+        exampleUsages = listOf("fun stuff", "airhorn"),
+        parameters = CommandParameters(
+            parameters = listOf(
+                CommandParameters.RequiredCommandParameter("tag-name"),
+                CommandParameters.RequiredCommandParameter("sound-name"),
+            ),
+        ),
+    ) {
+    override suspend fun execute(command: Command): CommandProcessingResult {
+        val args = command.rawArgs?.split('|')
```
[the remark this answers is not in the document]
DzikowskiW: seems like `command.getArgsByDelimiter("|")` does the trick, I will change it

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/TagAddCommandProcessor.kt
```diff
@@ -0,0 +1,38 @@
...
+    ) {
+    override suspend fun execute(command: Command): CommandProcessingResult {
+        val args = command.getArgsByDelimiter(this.parameters.delimiter)
+        if (args.size != 2) {
+            return error("You have to provide tag-name|sound-name in the arguments", logger)
+        }
+        val (tagName, soundName) = args
+
+        xSoundsService.addTagToXSound(soundName, tagName)
+            ?: return error("Sound $soundName doesn't exist", logger)
```
Deseteral, line 34: As it's implemented right now, you don't know that the sound wasn't added because it does not exist. You just know that is wasn't added.

`XSoundsService::addTagToXSound` should return Result type with information about the specific error. You might need help with this as it's currently only being used in `TextCommandParser`. @gorzalczany does this in his PR (#275).

Deseteral (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/domain/xsounds/XSoundsService.kt
```diff
@@ -45,4 +46,8 @@ class XSoundsService(private val repository: XSoundsRepository, private val file
             logger.error { "Could not increment play count for sound $soundName" }
         }
     }
+
+    suspend fun addTagToXSound(soundName: String, tag: String): Result<XSound, AddXTagToXSoundError> {
```
Deseteral, line 50: We shouldn't reuse the error type from repository in the service. Those should be separate types - repository error should be for database/storage related issues and service one should concern the business logic of the action.
In this case they will be the same, but this may not be true always.

Deseteral (approved):

Deseteral (approved):

PR #268, opened 2024-11-06
[beta backend] add song queue command

gorzalczany:


service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SongQueueCommandProcessor.kt
```diff
@@ -0,0 +1,62 @@
...
+        parameters = CommandParameters(
+            parameters = listOf(
+                CommandParameters.RequiredMultiValue("video-name", "youtube-id", "youtube-playlist-id"),
+            ),
+        ),
+    ) {
+    override suspend fun execute(command: Command): CommandProcessingResult {
+        val videoOrPlaylistId = command.rawArgs.orEmpty()
+        if (videoOrPlaylistId.isEmpty()) {
+            error("Missing exact song name or youtube's video/playlist id.", logger)
```
Deseteral, line 37: You're missing a return there
```suggestion
            return error("Missing exact song name or youtube's video/playlist id.", logger)
```

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SongQueueCommandProcessor.kt
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        val songs = mutableListOf<Song>()
+
+        songsService.getSongByNameWithYouTubeIdFallback(videoOrPlaylistId)?.let {
+            songs.add(it)
```
Deseteral, line 43: Why not just use song list returned by the service?

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SongQueueCommandProcessor.kt
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        // TODO: handle playlist
+
+        if (songs.isEmpty()) {
+            return error("Could not queue any song.", logger)
+        }
+
+        // TODO: filter only embeddable
```
Deseteral, line 52: This may not be worth doing if we're going to switch over to HTML5 video player.

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SongQueueCommandProcessor.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        eventStream.sendToEveryone(QueueSongsEvent(songs = songs))
```
Deseteral, line 54: ??
```suggestion
        eventStream.sendToEveryone(QueueSongsEvent(songs))
```

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
```diff
@@ -1,7 +1,36 @@
...
+
+        val maybeYoutubeId = nameOrYouTubeId.split(" ").firstOrNull()
+
+        return maybeYoutubeId?.let {
+            repository.findByYoutubeId(it) ?: createNewSong(it)
+        }
+    }
+
+    private suspend fun createNewSong(youtubeId: String, songName: String? = null): Song? {
+        val name = songName ?: youtubeClient.getVideoName(youtubeId).get()
```
Deseteral, line 27: Wouldn't this work?
```suggestion
        val name = songName ?: youtubeClient.getVideoName(youtubeId).get() ?: return null
```

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
```diff
@@ -1,7 +1,36 @@
...
+        if (name == null) {
+            return null
+        }
+
+        val newSong = Song(name, youtubeId, 0, null, null)
+        repository.insert(newSong)
+        return newSong
```
Deseteral, line 34: Make `insert` return what it inserted and then return it from here instead of `newSong`. They should be the same, except when the insert fails (returns `null`) which is an error you don't handle here.

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
```diff
@@ -1,7 +1,36 @@
...
+        }
+    }
+
+    private suspend fun createNewSong(youtubeId: String, songName: String? = null): Song? {
+        val name = songName ?: youtubeClient.getVideoName(youtubeId).get()
+        if (name == null) {
+            return null
+        }
+
+        val newSong = Song(name, youtubeId, 0, null, null)
```
Deseteral, line 32: I think named parameters would be more readable here.

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
```diff
@@ -1,7 +1,36 @@
...
-class SongsService(private val repository: SongsRepository) {
+import com.github.michaelbull.result.get
+import xyz.lebkuchenfm.external.youtube.YoutubeClient
+
+class SongsService(private val repository: SongsRepository, private val youtubeClient: YoutubeClient) {
     suspend fun getAllSongs(): List<Song> {
         return repository.findAllOrderByNameAsc()
     }
+
+    suspend fun incrementPlayCount(youtubeId: String): Song? {
```
Deseteral, line 11: Why `youtubeId`? Just use the `Song` instance, or at least `name`.

Deseteral (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
gorzalczany: When w use `q` and its argument as a name is not found in database then we are getting song from youtube and there id is const but name/title can be changed overtime. I thought it would be safer to use youtube id for that particular reason.

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
gorzalczany: I've added check if write was acknowledged but returning "newSong" object seems like desired behaviour 
https://stackoverflow.com/a/40780073/9865234

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SongQueueCommandProcessor.kt
```diff
@@ -0,0 +1,62 @@
...
+private val logger = KotlinLogging.logger {}
+
+class SongQueueCommandProcessor(private val songsService: SongsService, private val eventStream: EventStream) :
+    CommandProcessor(
+        key = "song-queue",
+        shortKey = "q",
+        helpMessage = """
+            Adds a song from the database to the queue, and if it is not there,
+            treats the phrase as a YouTube video ID or YouTube playlist ID.
+        """.trimIndent(),
```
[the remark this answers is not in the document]
gorzalczany: I think that breaking line here helps with message readability. I would leave it as is. Can be adjusted later.

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SongQueueCommandProcessor.kt
gorzalczany: I would leave it as is. It will be easier to see where we need to implement playlist handling.

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SongQueueCommandProcessor.kt
gorzalczany: I would leave that comment for now, and delete it later when HTML5 proves to be working as expected.

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
```diff
@@ -1,7 +1,36 @@
 package xyz.lebkuchenfm.domain.songs
 
-class SongsService(private val repository: SongsRepository) {
+import com.github.michaelbull.result.get
+import xyz.lebkuchenfm.external.youtube.YoutubeClient
+
+class SongsService(private val repository: SongsRepository, private val youtubeClient: YoutubeClient) {
```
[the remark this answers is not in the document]
gorzalczany: improved (I think)

PR #270, opened 2024-11-07
Deploy v3 beta

Deseteral:


beta.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ARG JDK_VERSION=23
+
+################################################################################
+# Stage 1: Create a stage for building frontend client.
+FROM node:${NODE_VERSION}-alpine AS client
+WORKDIR /usr/src/app
+
+COPY ./package.json .
+COPY ./yarn.lock .
+COPY ./packages/ ./packages
```
gorzalczany, line 11: should we have trailing / here? 🤔

beta.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+################################################################################
+# Stage 3: Build Application
+FROM gradle:latest AS build
+COPY --from=cache /home/gradle/cache_home /home/gradle/.gradle
+COPY ./service /usr/src/app/
+WORKDIR /usr/src/app
+COPY --chown=gradle:gradle ./service /home/gradle/src
+WORKDIR /home/gradle/src
+COPY --from=client /usr/src/app/packages/client-beta/dist/ ./src/main/resources/static
```
gorzalczany, line 33: comment above that line could be helpful ;)

beta.dockerfile
```diff
@@ -0,0 +1,44 @@
+ARG NODE_VERSION=18
+ARG JDK_VERSION=23
+
+################################################################################
+# Stage 1: Create a stage for building frontend client.
```
gorzalczany, line 5: Maybe we should adjust this stage description so that it fits with others

beta.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+COPY --chown=gradle:gradle ./service /home/gradle/src
+WORKDIR /home/gradle/src
+COPY --from=client /usr/src/app/packages/client-beta/dist/ ./src/main/resources/static
+# Build the fat JAR, Gradle also supports shadow
+# and boot JAR by default.
+RUN gradle buildFatJar --no-daemon
+
+################################################################################
+# Stage 4: Create the Runtime Image
+FROM eclipse-temurin:${JDK_VERSION} AS runtime
```
gorzalczany, line 40: is there a point in naming this stage

fly.beta.toml
```diff
@@ -0,0 +1,39 @@
+app = 'lebkuchen-fm-beta'
+primary_region = 'ewr'
```
gorzalczany, line 2: ewr? 🤔

gorzalczany (approved):

fly.beta.toml
Deseteral: Since the beginning of this year we're running in ewr region which is on the east coast of the USA. This is because Google Gemini is not available in the EU.

beta.dockerfile
Deseteral: I don't know, most of the dockerfile is copied from [Ktor docs](https://ktor.io/docs/docker.html#manual).

beta.dockerfile
Deseteral: Check out if this is what you mean.

beta.dockerfile
Deseteral: Done.

beta.dockerfile
Deseteral: We don't have to.

PR #273, opened 2024-11-09
Ktor WebSockets

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/Application.kt
```diff
@@ -117,13 +121,18 @@ fun Application.module() {
         }
     }
 
+    install(WebSockets) {
```
gorzalczany, line 124: I feel like we might want to move all those `install`s to separate method
- our own helper with all plugins
or
- move socket installation to eventStream route https://ktor.io/docs/server-plugins.html#install-route

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
```diff
@@ -0,0 +1,64 @@
...
+
+class Connection(val session: DefaultWebSocketSession)
+
+fun Route.eventStreamRouting(eventStream: WebSocketEventStream) {
+    webSocket("/event-stream") {
+        val connection = Connection(this)
+        eventStream.addConnection(connection)
+
+        for (frame in incoming) {
+            // TODO: This does not work yet, but I fix it when there are actual events coming from the client.
```
gorzalczany, line 27: Should we note it in the Doc?

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
```diff
@@ -0,0 +1,64 @@
...
+            val event = converter?.deserialize(call.request.headers.suitableCharset(), typeInfo<Event>(), frame)
+            println(event)
```
gorzalczany, line 29: replace println with logger?

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
```diff
@@ -0,0 +1,64 @@
...
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.ClassDiscriminatorMode
+import kotlinx.serialization.json.Json
+import xyz.lebkuchenfm.domain.eventstream.Event
+import xyz.lebkuchenfm.domain.eventstream.EventStream
+import java.util.Collections
+
+class Connection(val session: DefaultWebSocketSession)
+
+fun Route.eventStreamRouting(eventStream: WebSocketEventStream) {
```
gorzalczany, line 21: WDYT about moving this `fun` to new `EventStreamEndpoint` file and leave only `class WebSocketEventStream` here?

gorzalczany (dismissed):

service/src/main/kotlin/xyz/lebkuchenfm/Application.kt
Deseteral: I feel that it's better the way it's now. It's not even 150 lines long, and I find it more readable that the whole configuration is in single file.

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
Deseteral: No, that's what's necessary for WebSockets to function properly, and that's what I'm working on now.

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
Deseteral: No, this will be gone in the next PR.

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
Deseteral: Sure.

gorzalczany (approved):

PR #276, opened 2024-11-09
Pass execution context to the commands

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/ExecutionContext.kt
```diff
@@ -0,0 +1,5 @@
+package xyz.lebkuchenfm.domain.commands
+
+import xyz.lebkuchenfm.domain.auth.UserSession
+
+data class ExecutionContext(val session: UserSession)
```
gorzalczany, line 5: Idea for later (the Doc):
- add client context - is it discord command or web client interface. Store statistics of usage.

gorzalczany (approved):

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/ExecutionContext.kt
Deseteral: That's why I've made an `ExecutionContext` instead of just passing `UserSession`. If we find use for it, we will add what's necessary.

PR #280, opened 2024-11-13
[beta backend] add History repository

gorzalczany:


service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
```diff
@@ -1,14 +1,24 @@
...
+) {
     suspend fun getAllSongs(): List<Song> {
         return songsRepository.findAllOrderByNameAsc()
     }
 
-    suspend fun incrementPlayCount(song: Song): Song? {
-        return songsRepository.incrementPlayCountByName(song.name)
+    suspend fun incrementPlayCount(song: Song, userSession: UserSession): Song? {
+        return songsRepository.incrementPlayCountByName(song.name)?.also {
+            historyRepository.insert(HistoryEntry(Clock.System.now(), it.name, userSession.name))
```
Deseteral, line 20: We should get current time from a `CurrentDateTimeProvider` (or something like that), so that we can mock it in tests.
We can do that later though, if you don't want to make that change now.

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/HistoryMongoRepository.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    override suspend fun insert(history: HistoryEntry): Result<HistoryEntry, HistoryRepositoryError> {
+        return try {
+            if (collection.insertOne(history.toEntity()).wasAcknowledged()) {
+                Ok(history)
+            } else {
+                Err(HistoryRepositoryError.UnknownError)
+            }
+        } catch (e: MongoWriteException) {
+            Err((HistoryRepositoryError.UnknownError))
```
Deseteral, line 26:
```suggestion
            Err(HistoryRepositoryError.UnknownError)
```

Deseteral (dismissed):

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
gorzalczany: Noted in Docs about code testability.

Deseteral (approved):

PR #281, opened 2024-11-13
[beta backend] create xSoundsRepository unique index

gorzalczany:


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/XSoundsMongoRepository.kt
```diff
@@ -29,6 +32,12 @@ class XSoundsMongoRepository(database: MongoDatabase) : XSoundsRepository {
     private val collection = database.getCollection<XSoundEntity>("x")
     private val sortByName = Sorts.ascending(XSoundEntity::name.name)
 
+    suspend fun createUniqueIndex() {
+        val index = Indexes.ascending(XSoundEntity::name.name)
+        val options = IndexOptions().unique(true).name("unique_name")
+        collection.createIndex(index, options)
```
gorzalczany, line 38: In case where index could not be created it throws such exception that I think there is no point in catching it.
```

try {
    val indexOptions = IndexOptions().unique(true)
    val resultCreateIndex = theatersCollection.createIndex(
        Indexes.descending(Theater::theaterId.name), indexOptions
    )
    println("Index created: $resultCreateIndex")
} catch (e: DuplicateKeyException) {
    println("duplicate field values encountered, couldn't create index: \t${e.message}")
}
```

Deseteral (approved):

PR #288, opened 2024-11-17
Event stream improvements

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
```diff
@@ -6,31 +6,32 @@ import io.ktor.server.websocket.WebSocketServerSession
 import io.ktor.server.websocket.converter
 import xyz.lebkuchenfm.domain.eventstream.Event
 import xyz.lebkuchenfm.domain.eventstream.EventStream
-import java.util.Collections
-
-class Connection(val session: WebSocketServerSession)
+import xyz.lebkuchenfm.domain.eventstream.EventStreamConsumerId
+import java.util.UUID
```
gorzalczany, line 10: WDYT about using experimental kotlin uuid? (`@OptIn(ExperimentalUuidApi::class)`)

service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/EventStream.kt
```diff
@@ -1,5 +1,33 @@
...
+
+    open fun subscribe(consumer: ConsumerT) {
+        subscriptions[consumer.id] = consumer
+    }
+
+    open fun unsubscribe(consumer: ConsumerT) {
+        subscriptions.remove(consumer.id)
+    }
+
+    val subscriptionCount get() = subscriptions.count()
```
gorzalczany, line 26: do we need that?

service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/EventStream.kt
```diff
@@ -1,5 +1,33 @@
 package xyz.lebkuchenfm.domain.eventstream
 
-interface EventStream {
-    suspend fun sendToEveryone(event: Event)
+import io.ktor.util.collections.ConcurrentMap
+import java.util.UUID
+
+abstract class EventStream<ConsumerT : EventStream.Consumer> {
+    protected val subscriptions: MutableMap<EventStreamConsumerId, ConsumerT> = ConcurrentMap()
```
gorzalczany, line 7: `streamConsumers`?

gorzalczany (approved):

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/WebSocketEventStream.kt
Deseteral: I didn't want to use it because it's experimental, and was introduced only recently in Kotlin 2.0. I would prefer to wait for it to mature, and then swap it.

service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/EventStream.kt
Deseteral: Not in this branch, but it's used in my branch. I can remove it for now.

service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/EventStream.kt
Deseteral: I would prefer it as it is. You're subscribing to event stream, so this is a subscription list.

PR #293, opened 2024-11-22
redirect to login when something wrong with auth

mswiechowicz:


Deseteral:
Only 401 should redirect to login form.

PR #297, opened 2024-12-01
Reject modernity, embrace write concern

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/XSoundsMongoRepository.kt
```diff
@@ -57,19 +60,14 @@ class XSoundsMongoRepository(database: MongoDatabase) : XSoundsRepository {
...
-        } catch (e: MongoWriteException) {
-            if (e.error.category == ErrorCategory.DUPLICATE_KEY) {
-                Err(XSoundsRepositoryError.SoundAlreadyExists)
-            } else {
-                Err((XSoundsRepositoryError.UnknownError))
+        return runSuspendCatching { collection.insertOne(sound.toEntity()) }
+            .map { sound }
+            .mapError { ex ->
+                when {
+                    ex is MongoWriteException && ex.isDuplicateKeyException -> XSoundsRepositoryError.SoundAlreadyExists
```
gorzalczany, line 67: I prefer shorter conditions in when. 
```kotlin
.mapError { ex ->
    val isWritingDuplicate = ex is MongoWriteException && ex.isDuplicateKeyException
    
    when {
        isWritingDuplicate -> XSoundsRepositoryError.SoundAlreadyExists
        else -> XSoundsRepositoryError.UnknownError
    }
}
```

service/src/main/kotlin/xyz/lebkuchenfm/domain/history/HistoryRepository.kt
```diff
@@ -3,9 +3,9 @@ package xyz.lebkuchenfm.domain.history
 import com.github.michaelbull.result.Result
 
 interface HistoryRepository {
-    suspend fun insert(history: HistoryEntry): Result<HistoryEntry, HistoryRepositoryError>
+    suspend fun insert(history: HistoryEntry): Result<HistoryEntry, InsertHistoryEntryError>
 }
 
-sealed class HistoryRepositoryError {
-    data object UnknownError : HistoryRepositoryError()
+sealed class InsertHistoryEntryError {
```
gorzalczany, line 9: WDYT about not specifying it to `Insert` Error?
```kotlin
sealed class HistoryRepositoryError {
     data object WritingHistoryError : HistoryRepositoryError()
     // because soon we might want to
     data object ReadingHistoryError : HistoryRepositoryError()
 }
 ```

File: service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/SongsMongoRepository.kt [hunk 3]
```diff
@@ -25,8 +30,10 @@ class SongsMongoRepository(database: MongoDatabase) : SongsRepository {
         collection.createIndex(Indexes.text(SongEntity::name.name), options)
     }
 
-    override suspend fun insert(song: Song): Boolean {
-        return collection.insertOne(SongEntity(song)).wasAcknowledged()
+    override suspend fun insert(song: Song): Result<Song, InsertSongError> {
+        return runSuspendCatching { collection.insertOne(song.toEntity()) }
+            .map { song }
+            .mapError { InsertSongError.UnknownError }
```
gorzalczany, line 36: that looks awesome 👏🏻

gorzalczany (approved):
.kt part LGTM

service/src/main/kotlin/xyz/lebkuchenfm/domain/history/HistoryRepository.kt
Deseteral: No, each operation should have its own error type. If we don't do that we will make matching on error type useless (why do we need do handle writing error, when we're reading?).

PR #302, opened 2024-12-07
Create user session

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/api/commands/CommandsEndpoint.kt
```diff
@@ -14,10 +16,8 @@ private val logger = KotlinLogging.logger {}
 
 fun Route.commandsRouting(commandExecutorService: CommandExecutorService) {
     post("/commands/text") {
-        // TODO: Get UserSession from request when the endpoints get authorization.
-        val session = UserSession("FAKE USER TODO")
-        // val session = call.sessions.get<UserSession>()
-        // checkNotNull(session)
+        val session = call.sessions.get<UserSession>()
+        checkNotNull(session)
```
gorzalczany, line 20: Don't we want to catch it and produce user friendly response?

service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
```diff
@@ -53,13 +59,41 @@ class DiscordClient(config: ApplicationConfig, private val commandExecutorServic
...
-                    content = result.message.markdown
+                val author = requireNotNull(it.author)
+                val user = userService.getByDiscordId(author.id.toString())
+
+                // TODO: Add support for login command.
+                val isLoginCommand = false
+
+                when {
+                    isLoginCommand -> {
+                        val context = ExecutionContext(UserSession(name = "unknown"))
```
gorzalczany, line 73: Shouldn't session in context be optional for such a case?

service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
```diff
@@ -53,13 +59,41 @@ class DiscordClient(config: ApplicationConfig, private val commandExecutorServic
...
+                            content = result.message.markdown
+                        }
+                    }
+
+                    else -> {
+                        // TODO: Add support for login command.
+                        it.reply {
+                            content = """
+                                You have to connect your Discord account with LebkuchenFM.
+                                Use `$commandPrompt TODO LOGIN_COMMAND_KEY <lebkuchen-fm-username>` to login.
```
gorzalczany, line 93: shouldn't we just pass command to execution service and allow it to response with help message? 🤔

service/src/main/kotlin/xyz/lebkuchenfm/api/commands/CommandsEndpoint.kt
Deseteral: Those endpoints are behind auth strategy. There will always be session here. So realistically `checkNotNull` will never throw. 
If it does a developer made a mistake and that's illegal state, that is not presentable to the end user - it's just internal server error. That will never happen.

service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
Deseteral: But this is only the case for login command and it does not use session data. Making it optional will make us do useless not null check all over the code because of this.
I'm personally more interested in changing the Discord auth flow entirely (removing login command), and making this field optional is not worth it in my opinion.

service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
Deseteral: I don't think so. This specific message is a 403 of sorts for Discord. Your user is not assigned to any LebkuchenFM user, so you're basically not logged in.

service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
gorzalczany: k, but having "unknown" passed as a name on production smells 💩

gorzalczany: also write down to the doc that part about rewriting discord auth flow ;)

gorzalczany (dismissed):

service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
Deseteral: We agreed not to do it that way. We will worry about login command when we actually implement it.

gorzalczany (approved):

PR #305, opened 2024-12-10
add playilists support to `queue` command

gorzalczany:
### What:
- add playlist support

### Why:
- we want to queue Klocuch's Christmas carols


service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
```diff
@@ -63,4 +70,16 @@ class SongsService(
...
+    private suspend fun getOrCreateSongsCorrespondingToYoutubeVideos(youtubeVideos: List<YoutubeVideo>): List<Song> {
+        val youtubeIds = youtubeVideos.map { it.id }
+        val knownSongs = songsRepository.findByYoutubeIds(youtubeIds)
+        val knownIds = knownSongs.map { it.youtubeId }
+
+        val newSongs = youtubeVideos
+            .filterNot { it.id in knownIds }
+            .mapNotNull { createNewSong(it.id, it.name) }
+
+        return knownSongs + newSongs
```
Deseteral, line 83: This way we're losing sorting from original playlist.

Deseteral (dismissed):

service/src/main/kotlin/xyz/lebkuchenfm/domain/songs/SongsService.kt
gorzalczany: very good catch!

Deseteral (approved):

PR #317, opened 2025-01-12
Soundboard should actually play some sounds

Deseteral:


packages/client-beta/src/apps/Player/services/player-commands.ts
Deseteral, line 1: Dead code. None of the methods are currently used and we should not use commands like that anymore. We should have dedicated endpoints, with correct error handling for these kinds of actions.

service/src/main/kotlin/xyz/lebkuchenfm/domain/soundboard/SoundboardService.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.github.michaelbull.result.Result
+import xyz.lebkuchenfm.domain.eventstream.Event
+import xyz.lebkuchenfm.domain.eventstream.EventStream
+import xyz.lebkuchenfm.domain.xsounds.XSoundsService
+
+class SoundboardService(
+    private val xSoundsService: XSoundsService,
+    private val eventStream: EventStream<*>,
+) {
+    suspend fun playXSound(soundName: String): Result<Unit, PlayXSoundError> {
```
gorzalczany, line 14: maybe we should return xSound on success? Dunno why but I have mixed feelings about Unit here

packages/client-beta/src/apps/Soundboard/Soundboard.tsx
```diff
@@ -25,8 +26,12 @@ function Soundboard() {
     }
   };
 
-  function playXSoundLocally(url: string) {
-    XSoundsPlayService.play(url);
+  function playXSound(sound: XSound, event: MouseEvent) {
+    if (event.metaKey || event.altKey) {
+      XSoundsPlayService.play(sound.url);
```
gorzalczany, line 31: WDYT about removing `XSoundsPlayService` and moving its logic to `XSoundsPlayService` as `SoundboardService.playXSoundLocally` function?

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/XCommandProcessor.kt
```diff
@@ -28,12 +29,13 @@ class XCommandProcessor(private val xSoundsService: XSoundsService, private val
...
 
-        val xSound = xSoundsService.getByName(soundName)
-            ?: return error("Sound '$soundName' does not exist.", logger)
-
-        eventStream.sendToEveryone(Event.PlayXSound(soundUrl = xSound.url))
-        xSoundsService.markAsPlayed(xSound.name)
-
-        return CommandProcessingResult.fromMarkdown("Played $soundName sound.")
+        return soundboardService.playXSound(soundName)
+            .map { CommandProcessingResult.fromMarkdown("Played $soundName sound.") }
```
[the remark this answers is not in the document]
Deseteral: Actually `mapBoth` works great here. Didn't know it existed until now :).

service/src/main/kotlin/xyz/lebkuchenfm/domain/soundboard/SoundboardService.kt
Deseteral: Unit is perfectly fine here. We just want to now that the operation succeeded, but it does not have any value associated with its success.

packages/client-beta/src/apps/Soundboard/Soundboard.tsx
Deseteral: I'm not sure. `XSoundsPlayService` is all about reacting to the events and playing the sound locally. I'm not sure if this is the right place for actually playing the soundboard sound.

gorzalczany: So naming became a bit confusing. Why `XSoundsPlayService` is about locally playing any url? 
For me me it should be reversed. `XSoundsPlayService` sends request to backend, and `SoundsService` plays locally any sound (not only x?) from provided url.

Deseteral: I see. I will try to propose something that makes more sense.

gorzalczany (approved):

packages/client-beta/src/services/play-x-sound-event-handler.ts
```diff
@@ -0,0 +1,33 @@
+import { UserPreferencesService } from './user-preferences-service';
```
kstodolak, line 1: Nie rozumiem, dlaczego ten mój poprzedni plik jest usunięty. On robi prawie to samo co ten Twój nowy. Mamy teraz 3 "serwisy", które nazwą wskazują, że robią coś z dźwiękiem. Na pierwszy rzut oka mocno to się pokomplikowało. Dla mnie puszczanie dźwięku lokalnie brzmi tak, że może być w serwisie od puszczania xSounds i tak był nazwany poprzedni plik.

packages/client-beta/src/apps/Soundboard/soundboard-service.ts
```diff
@@ -0,0 +1,45 @@
+import { XSound } from '../../types/x-sound';
```
kstodolak, line 1: Proponuje przenieść do folderu `./services` tak jak serwisy dla aplikacji Player

kstodolak (changes requested):

kstodolak (approved):

PR #318, opened 2025-01-12
Simplify error responses with respondWithProblem

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/api/auth/ValidateAuthHandler.kt
```diff
@@ -18,74 +15,73 @@ import xyz.lebkuchenfm.domain.users.SetPasswordError
...
-                        "Could not authenticate.",
-                        "Wrong password was provided.",
-                        status,
-                        instance,
-                    )
+                when (error) {
+                    AuthError.BadCredentialsError -> {
+                        call.respondWithProblem(
+                            title = "Could not authenticate.",
+                            detail = "Wrong password was provided.",
```
gorzalczany, line 25: `Wrong password was provided.`
or 
`Wrong username or password was provided.`

we should not indicate that username exists when user typed wrong credentials

service/src/main/kotlin/xyz/lebkuchenfm/api/auth/ValidateAuthHandler.kt
```diff
@@ -18,74 +15,73 @@ import xyz.lebkuchenfm.domain.users.SetPasswordError
...
-                    AuthError.UserDoesNotExistError -> ProblemResponse(
-                        "Could not authenticate.",
-                        "User does not exist.",
-                        status,
-                        instance,
-                    )
+                    AuthError.UserDoesNotExistError -> {
+                        call.respondWithProblem(
+                            title = "Could not authenticate.",
+                            detail = "User does not exist.",
```
gorzalczany, line 33: same as above?

service/src/main/kotlin/xyz/lebkuchenfm/api/auth/ValidateAuthHandler.kt
```diff
@@ -18,74 +15,73 @@ import xyz.lebkuchenfm.domain.users.SetPasswordError
...
-                            ).joinToString(separator = " ")
-                            SetPasswordError.UnknownError -> "Something went wrong."
-                        }
-                        ProblemResponse(title, detail, status, instance)
+                        call.respondWithProblem(
+                            title = "Could not set password for user.",
+                            detail = when (error.setPasswordError) {
+                                SetPasswordError.UserDoesNotExist -> "User does not exist."
+                                is SetPasswordError.ValidationError -> listOfNotNull(
+                                    "Password is too weak:",
```
gorzalczany, line 55: I truly dislike any password requirements.
IMO one and only requirement should be that it must exists so `len(password) > 0`

service/src/main/kotlin/xyz/lebkuchenfm/api/auth/ValidateAuthHandler.kt
Deseteral: Fixed

service/src/main/kotlin/xyz/lebkuchenfm/api/auth/ValidateAuthHandler.kt
Deseteral: I don't agree.

gorzalczany (approved):

PR #347, opened 2025-02-13
Deploy v3 episode 4: Update readme

Deseteral:


docs/configuration.md
```diff
@@ -0,0 +1,45 @@
+# Configuration
+LebkuchenFM uses environment variables for configuration.
+
+You can also create `.env` file in the root of this project and put desired configuration variables in it.
```
gorzalczany, line 4: It is not so straightforward now. Backend service does not read this automatically. Maybe we should describe this with recommended IDE as an example. Or even describe `.env` files as recommended solution but with "some caveats" - .env plugin in intelij, docker compose env file vs plain properties in yaml? 🤔

docs/configuration.md
```diff
@@ -0,0 +1,45 @@
...
+## YouTube player
+YouTube integration is optional - when required variables are not provided YouTube functionalities are not available.
+
+| Name              | Description             | Required | Default value |
+|-------------------|-------------------------|----------|---------------|
+| `YOUTUBE_API_KEY` | YouTube Data API token. | Yes      | -             |
+
+## Discord
+Discord integration is optional - when required variables are not provided Discord communication is not available.
+You can still issue commands using the [LebkuchenFM REST API](./rest_api.md).
```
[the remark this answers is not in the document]
Deseteral: I forgot to remove the link.

docs/configuration.md
Deseteral: That's very true. I totally forgot about it.

docs/development.md
```diff
@@ -7,13 +7,16 @@ Refer to [Ktor documentation](https://ktor.io/docs) for detailed building instru
 ## Running locally
 
 ### Configuration
-Create `.env` file in the root of this project and put desired configuration variables in it.
+LebkuchenFM is configured via environment variables.
 Refer to [configuration](./configuration.md) documentation for available configuration variables.
 
+We recommend using `.env` file to store configuration variables for local development.
```
gorzalczany, line 13: maybe this paragraph would be of better use under `Configuration` in docs/configuration.md

gorzalczany (approved):

docs/development.md
Deseteral: No, I don't think so. Recommending `.env` files only makes sense for development. When you're deploying you have to adjust to what your server provider is forcing you to use (fly.io admin vs maybe `.env` files when self-hosting).

PR #348, opened 2025-02-13
Deploy v4 episode 5: Set project version

Deseteral:


scripts/set_version.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/sh
+
+# Set the version in Gradle project.
+sed -i '' '1s/.*/lebkuchenfm_version='"$1"'/' ./service/gradle.properties
```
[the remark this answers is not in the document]
Deseteral: Build will fail if we don't.

.github/workflows/version-check.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  check-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check version equality between modules
+        run: |
+          GRADLE_VERSION=$(awk -F= 'NR==1 {print $2}' ./service/gradle.properties)
+          NPM_VERSION=$(jq -r '.version' ./client/package.json)
```
gorzalczany, line 17: are we sure jq is available on GitHub runners?

.github/workflows/version-check.yml
gorzalczany, line 1: ❤️

scripts/set_version.sh
gorzalczany, line 1: if PR with Readme will be merged first then this file should be described in `scripts` section.

Deseteral: That was my intention :)

.github/workflows/version-check.yml
Deseteral: Well the thing passes so it must be :) I wasn't sure at first, but it works.

docs/scripts.md
```diff
@@ -1,25 +1,35 @@
...
 ## FM command helper
-Helper script `scripts/fm.sh` is available for local development purposes. By default, it sends a command to a local development server.
+Helper script `scripts/fm` is available for local development purposes. By default, it sends a command to a local development server.
 
 Example commands:
 ```
-> fm.sh "/fm resume" "user-api-token"
-> fm.sh "/fm q dQw4w9WgXcQ" "user-api-token"
-> fm.sh "/fm x alert" "user-api-token"
+$ scripts/fm "/fm resume" "user-api-token"
```
gorzalczany, line 17: be consistent about `$` at scripts, also this code block could be marked as sh

gorzalczany (approved):

scripts/fm
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /bin/sh
```
gorzalczany, line 1: I believe `bash` was better here. What if someones default shell is `fish` syntax would not match

gorzalczany (approved):

PR #354, opened 2025-02-27
LLM prompts

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/LlmPersonalityPromptsMongoRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.bson.types.ObjectId
+import xyz.lebkuchenfm.domain.radiopersonality.llmprompts.LlmPersonalityPrompt
+import xyz.lebkuchenfm.domain.radiopersonality.llmprompts.LlmPersonalityPromptsRepository
+import xyz.lebkuchenfm.domain.radiopersonality.llmprompts.LlmPromptCreation
+import xyz.lebkuchenfm.domain.users.UserName
+
+class LlmPersonalityPromptsMongoRepository(database: MongoDatabase) : LlmPersonalityPromptsRepository {
+    private val collection = database.getCollection<LlmPersonalityPromptEntity>("llmPersonalityPrompts")
+
+    override suspend fun findLatestActiveGroupedByName(): List<LlmPersonalityPrompt> {
```
gorzalczany, line 23: 🟢 Dunno why but is hard for me to conclude what does it do at first sight when reading that name.
But I could not came with anything better :/

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/llmprompts/LlmPromptService.kt
```diff
@@ -0,0 +1,24 @@
...
+            ?.text
+            ?: return null
+
+        val situationPromptText = situationPromptsRepository.findLatestByType(situationType)
+            ?.text
+            ?: return null
+
+        val fullPrompt = buildString {
+            append(personalityPromptText)
+            append(situationPromptText)
```
gorzalczany, line 19: 🟡 appendLine instead of append?

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/llmprompts/LlmPromptService.kt
```diff
@@ -0,0 +1,24 @@
...
+
+class LlmPromptService(
+    private val personalityPromptsRepository: LlmPersonalityPromptsRepository,
+    private val situationPromptsRepository: LlmSituationPromptsRepository,
+) {
+    suspend fun getFullPromptTextForSituationUsingRandomPersonality(situationType: LlmSituationType): String? {
+        val personalityPromptText = personalityPromptsRepository.findLatestActiveGroupedByName()
+            .randomOrNull()
+            ?.text
+            ?: return null
```
gorzalczany, line 11: 🔴 result with error `missing personality prompt`?

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/llmprompts/LlmPromptService.kt
```diff
@@ -0,0 +1,24 @@
...
+
+        val situationPromptText = situationPromptsRepository.findLatestByType(situationType)
+            ?.text
+            ?: return null
```
gorzalczany, line 15: 🔴 result with error `missing situational prompt`?

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/llmprompts/LlmSituationPromptsRepository.kt
```diff
@@ -0,0 +1,5 @@
+package xyz.lebkuchenfm.domain.radiopersonality.llmprompts
+
+interface LlmSituationPromptsRepository {
+    suspend fun findLatestByType(type: LlmSituationType): LlmSituationPrompt?
```
gorzalczany, line 4: 🤔 maybe we should allow multiple active situational prompts? 
that way Radio Presenter could  randomly "greet me" / "complain about the taxi driver" / "express happiness cause of holidays" etc instead of mixing all the above in one message

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/LlmSituationPromptsMongoRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+            .sort(Sorts.descending(LlmSituationPromptEntity::createdAt.name))
+            .limit(1)
+            .firstOrNull()
+            ?.toDomain()
+    }
+}
+
+@Serializable
+private data class LlmSituationPromptEntity(
+    @Contextual @SerialName("_id") val id: ObjectId? = null,
```
gorzalczany, line 33: 🤔  do we even need that id here?

gorzalczany (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/LlmSituationPromptsMongoRepository.kt
Deseteral: We don't. I was wondering about this one too. In many Mongo entities we have `id` field, even though we don't ever use it. I've added this because that's what that document actually looks like in Mongo. If we want to remove the `id` here, then we should also remove it everywhere else.

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/LlmPersonalityPromptsMongoRepository.kt
Deseteral: Same for me unfortunately. I'm also not happy about this name, but that's what it actually does , and I couldn't come up with anything better :/

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/llmprompts/LlmPromptService.kt
Deseteral: Sure, why not.

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/LlmSituationPromptsMongoRepository.kt
gorzalczany: +1 for removing everywhere

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/llmprompts/LlmPromptService.kt
Deseteral: I would argue that the return type here is correct but what we're missing is logging.
The return type here is correct because it's representing the absence of value, rather than an error.

That does not matter much though, because it turns out I could feed a system prompt to Gemini, so we need those calls separate after all.

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/llmprompts/LlmSituationPromptsRepository.kt
Deseteral: This should be expressed via personality prompts. Here we just need a description of what's happening - currently "song is changing" or "listener is calling". The whole sentiment of "you like to complain about the taxi drivers", "you love xmas" should be a part of personality prompt.

gorzalczany (approved):

PR #360, opened 2025-03-02
Make MongoDB entity declarations consistent

### LINKED ISSUE #359: Improve consistency of MongoDB entries serialization
Every MongoDB entry data class should have the same method of serialization.

Deseteral:
Closes #359.

### DIFF
--- a/.github/workflows/version-check.yml
+++ b/.github/workflows/version-check.yml
@@ -27,10 +27,7 @@ jobs:
       - name: Check version equality between branches
         run: |
           WORK_VERSION=$(awk -F= 'NR==1 {print $2}' ./service/gradle.properties)
-
-          git fetch origin main
-          git checkout main
-          MAIN_VERSION=$(awk -F= 'NR==1 {print $2}' ./service/gradle.properties)
+          MAIN_VERSION=$(git show main:service/gradle.properties | awk -F= 'NR==1 {print $2}')
 
           echo "Gradle version on working branch: '$WORK_VERSION'."
           echo "Gradle version on main branch:    '$MAIN_VERSION'."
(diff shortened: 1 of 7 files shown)


.github/workflows/version-check.yml
```diff
@@ -15,16 +15,27 @@ jobs:
...
             echo "Version mismatch between client and service. Update the version in both modules."
             exit 1
           fi
 
       - name: Check version equality between branches
         run: |
+          WORK_VERSION=$(awk -F= 'NR==1 {print $2}' ./service/gradle.properties)
+
           git fetch origin main
+          git checkout main
```
Deseteral, line 32: This is what you get for using AI to generate code and not checking if it even works.

.github/workflows/version-check.yml
```diff
@@ -15,16 +15,27 @@ jobs:
...
           MAIN_VERSION=$(awk -F= 'NR==1 {print $2}' ./service/gradle.properties)
-          if [ "$MAIN_VERSION" == "$GRADLE_VERSION" ]; then
+
+          echo "Gradle version on working branch: '$WORK_VERSION'."
+          echo "Gradle version on main branch:    '$MAIN_VERSION'."
+
+          if [ "$WORK_VERSION" == "$MAIN_VERSION" ]; then
             echo "Version number has not changed. Update the version before merging."
             exit 1
           fi
```
Deseteral, line 41: This is what you get for using AI to generate code and not checking if it even works.

.github/workflows/version-check.yml
gorzalczany, line 32: that checkout feels strange, but If it is in this workflow context then probably that's not a problem

gorzalczany (dismissed):

gorzalczany (approved):

PR #362, opened 2025-03-02
Say command handled by radio personality service

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/Event.kt
```diff
@@ -50,5 +52,10 @@ sealed interface Event {
     data class Say(
         val text: String,
         val audio: Base64EncodedAudio,
-    ) : Event
+    ) : Event {
+        companion object {
```
gorzalczany, line 56: im not sure if I like this here. I mean I like how it looks in commands processors but...

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/SayCommandProcessor.kt
```diff
@@ -8,13 +8,13 @@ import xyz.lebkuchenfm.domain.commands.model.Command
 import xyz.lebkuchenfm.domain.commands.model.CommandProcessingResult
 import xyz.lebkuchenfm.domain.eventstream.Event
 import xyz.lebkuchenfm.domain.eventstream.EventStream
-import xyz.lebkuchenfm.domain.radiopersonality.speechsynthesis.TextToSpeechProvider
```
gorzalczany, line 11: I wish there was `package private` visibility modifier to now hide `speechsynthesis` from the rest of the code

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/RadioPersonalityService.kt
```diff
@@ -38,6 +38,16 @@ class RadioPersonalityService(
 
         return RadioPersonalityUtterance(text, audio)
     }
+
+    suspend fun userControlledUtterance(userMessage: String): RadioPersonalityUtterance? {
+        val audio = textToSpeechProvider.synthesize(userMessage)
+            ?: run {
+                logger.error { "Could not synthesize the text to audio." }
```
gorzalczany, line 45: now we log almost same error twice: here and in command processor

service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/Event.kt
gorzalczany: Maybe we should create `RadioPersonalityUtterance+SayEvent.kt` file in `xyz.lebkuchenfm.domain.radiopersonality` package 

```
package xyz.lebkuchenfm.domain.radiopersonality

import xyz.lebkuchenfm.domain.eventstream.Event


fun RadioPersonalityUtterance.toSayEvent() = Event.Say(this.text, this.audio)

```

and then use it like 

```
eventStream.sendToEveryone(utterance.ToSayEvent())
```

🤔 

syntax is a bit worse, but separation feels better

gorzalczany (approved):

service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/Event.kt
Deseteral: Separation from what? We're creating an event. This method is just providing another way to create it. It's not really a radio personality thing, or command processor thing or even event stream thing. This is just a part of our domain.

service/src/main/kotlin/xyz/lebkuchenfm/domain/radiopersonality/RadioPersonalityService.kt
Deseteral: That's not true. Here we provide specific reason for the error. In command processor we have generic message that we're sending back to Discord.

PR #365, opened 2025-03-04
ElevenLabs TTS integration

Deseteral:


service/src/main/kotlin/xyz/lebkuchenfm/Application.kt
```diff
@@ -114,7 +116,20 @@ fun Application.module() {
     val llmSituationsPromptRepository = LlmSituationPromptsMongoRepository(database)
     val llmPromptService = LlmPromptService(llmPersonalityPromptRepository, llmSituationsPromptRepository)
 
-    val textToSpeechProvider = GoogleCloudTextToSpeech(googleCloudTextToSpeechClient)
+    val ttsProviderConfig = environment.config.property("radioPersonality.textToSpeech.ttsProvider").getString()
+    val textToSpeechProvider: TextToSpeechProvider = when (ttsProviderConfig) {
+        "googleCloud" -> {
```
gorzalczany, line 121: maybe we could have switch for that in web client?

service/src/main/kotlin/xyz/lebkuchenfm/external/elevenlabs/ElevenLabsClient.kt
```diff
@@ -0,0 +1,121 @@
...
+        val requestBody = TextSynthesizeRequestBody(
+            text = text,
+            modelId = "eleven_multilingual_v2",
+            voiceSettings = TextSynthesizeRequestBody.VoiceSettings(
+                speed = voiceSpeed,
+                stability = 0.5,
+                similarityBoost = 0.75,
+            ),
+        )
+        val format = "mp3_44100_128"
```
gorzalczany, line 64: wouldn't 96 be enough?

gorzalczany (approved):

service/src/main/kotlin/xyz/lebkuchenfm/Application.kt
Deseteral: We should, but right now we cannot! We need a system for changing the configuration during runtime. This is definitely something that's on my mind recently, but it's not really doable in the current version without some stupid hacks.

gorzalczany (approved):

Issue #325, opened 2025-02-07
New x-sound form does not handle encoding.


gorzalczany:
I tried to add sound with polish letter in the name and it did not work in such a case.
Removing polish characters resulted in successful sound upload.

- there was no error on that (that should be fixed in this PR)
- progress for upload would be a nice touch

gorzalczany:
We should extract 
> progress for upload would be a nice touch

to separate task

PR #366, opened 2025-03-05
[Bugfix] dropbox request encoding

gorzalczany:
### What:
- escape non ascii characters in `Dropbox-API-Arg`

### Why
- https://www.dropbox.com/developers/reference/json-encoding 

-----
Closes https://github.com/Deseteral/lebkuchen-fm/issues/325

### DIFF
--- a/service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxClient.kt
+++ b/service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxClient.kt
@@ -20,12 +20,14 @@ import io.ktor.client.request.post
 import io.ktor.client.request.setBody
 import io.ktor.client.request.url
 import io.ktor.http.ContentType
+import io.ktor.http.HeadersBuilder
 import io.ktor.http.URLBuilder
 import io.ktor.http.contentType
 import io.ktor.http.isSuccess
 import io.ktor.http.parameters
 import io.ktor.serialization.kotlinx.json.json
 import io.ktor.server.config.ApplicationConfig
+import kotlinx.serialization.KSerializer
 import kotlinx.serialization.json.Json
 import xyz.lebkuchenfm.external.storage.dropbox.models.DropboxFileSharing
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxClient.kt
```diff
@@ -162,4 +163,21 @@ class DropboxClient(config: ApplicationConfig) {
         data object ErrorWhenUploadingFile : DropboxClientError()
         data object ErrorWhenCreatingFileUrl : DropboxClientError()
     }
+
+    private fun <T> HeadersBuilder.appendDropboxApiArg(value: T, serializer: KSerializer<T>) {
+        // https://www.dropbox.com/developers/reference/json-encoding
+        fun String.encodedForDropbox() = buildString {
+            for (char in this@encodedForDropbox) {
+                if (char.code in 32..<127) {
```
Deseteral, line 171: Since the root of the problem is passing stuff in headers, what we actually need is just URL encoding like this:
```
URLEncoder.encode(value, Charsets.UTF_8)
```
That's also what the documentations says:
> If you use the arg URL parameter, generate a JSON string, encode it as UTF-8, then use the application/x-www-form-urlencoded rules to encode the value.

Deseteral (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxClient.kt
gorzalczany: I tried that. But then dropbox responded with http 400 and info that it expected `Content-Type: application/octet-stream`

gorzalczany: nvm, must have had a stroke yesterday or sth.

done

Deseteral (approved):

Issue #371, opened 2025-03-07
Button component


Deseteral:
A button. You click it, it does something.

PR #376, opened 2025-03-19
Simple playground for testing components. Button component updated.

### LINKED ISSUE #370: Place to see all component and their usage example
Something like:
- Storybook.
- Dedicated app in our OS that does same thing as Storybook.

kstodolak:
In `/client` directory you can run `yarn run playground` for new vite dev server only for testing component purposes.
image

Closes #370 #371

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@solidjs/router": "0.13.0",
+    "@types/node": "22.13.10",
     "events": "3.3.0",
     "mitt": "3.0.1",
     "solid-js": "1.8.15",
--- a/client/package.json
+++ b/client/package.json
@@ -11,7 +11,8 @@
     "format": "prettier --write .",
     "format:check": "prettier --check .",
     "preview": "vite preview",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --noEmit",
+    "playground": "vite --config ./vite.config.playground.ts"
   },
   "dependencies": {
     "@solidjs/router": "0.13.0",
@@ -22,6 +23,7 @@
   },
   "devDependencies": {
     "@types/yt-player": "3.5.4",
+    "@types/node": "22.13.10",
     "@typescript-eslint/eslint-plugin": "7.7.0",
     "@typescript-eslint/parser": "7.1.1",
     "clsx": "2.1.0",
--- a/client/package.json
+++ b/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "lebkuchen-fm-client",
   "private": true,
-  "version": "2025.2.0",
+  "version": "2025.3.0",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/client/package.json
+++ b/client/package.json
@@ -23,7 +23,6 @@
   },
   "devDependencies": {
     "@types/yt-player": "3.5.4",
-    "@types/node": "22.13.10",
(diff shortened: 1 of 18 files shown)


client/src/_playground/Buttons.playground.tsx
```diff
@@ -0,0 +1,39 @@
+import { Button, ButtonVariant } from '@components/Button/Button';
+
+export const ButtonsPlayground = () => (
```
Deseteral, line 3: Why arrow function instead of regular one?

client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "lebkuchen-fm-client",
   "private": true,
-  "version": "2025.2.0",
+  "version": "2025.3.0",
```
Deseteral, line 4: I wonder if it's major or minor version 🤔 @gorzalczany what do you think? `2025.3.0` or `2025.2.1`?

client/src/_playground/styles.playground.module.css
```diff
@@ -0,0 +1,18 @@
+.componentsPlayground {
+  background-color: #f1f1f1;
```
Deseteral, line 2: Background color could be a part of CSS theme perhaps?

client/vite.config.playground.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConfig } from 'vite';
+import solid from 'vite-plugin-solid';
+// @ts-ignore
```
Deseteral, line 3: Question - don't you want to add node typings?

client/vite.config.playground.ts
```diff
@@ -0,0 +1,23 @@
...
+    outDir: path.join(__dirname, 'src', '_playground', 'dist'),
+    rollupOptions: {
+      input: {
+        app: path.join(__dirname, 'src', '_playground', 'index.html'),
+      },
+    },
+  },
+  plugins: [solid()],
+  resolve: {
+    dedupe: ['react', 'react-dom'],
```
Deseteral, line 18: We don't use react, so why do we need that?

Deseteral (changes requested):

client/src/_playground/Buttons.playground.tsx
kstodolak: Just a playground - I didn't pay much attention to it. I can change it :)

client/src/_playground/styles.playground.module.css
kstodolak: Yes it can be. I just set random, neutral color for a background for the tested components. TBH I treated this playground like PoC to show you this and take feedback from you all. But I will change it to already existing color variable :D

client/vite.config.playground.ts
kstodolak: Just reused most of the configuration from the existing config. Good question I am gonna check it

client/package.json
gorzalczany: Hard question. It would be a really good use case for feature/release branch, but at the same time there is no point in freezing this code. I think that it is a feature, even if it is "just" a feature for developers so 2025.3.0 sounds good to me.

Deseteral: `2025.3.0` it is then 👍

client/src/_playground/styles.playground.module.css
Deseteral: I would be good to see all of those components on the background color we use everywhere else (in apps).

Deseteral (approved):

PR #389, opened 2025-05-31
Bugfix | dropbox auth

gorzalczany:
attempt to fix dropbox auth bug
also add more logs for further investigation

### DIFF
--- a/service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxClient.kt
+++ b/service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxClient.kt
@@ -117,11 +117,19 @@ class DropboxClient(config: ApplicationConfig) {
                 }
                 sendWithoutRequest { true }
                 refreshTokens {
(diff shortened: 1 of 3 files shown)


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/dropbox/DropboxClient.kt
```diff
@@ -117,11 +117,19 @@ class DropboxClient(config: ApplicationConfig) {
                 }
                 sendWithoutRequest { true }
                 refreshTokens {
-                    val tokenInfo: DropboxTokenInfo = client.post(tokenPostRequest).body()
-                    bearerTokenStorage.clear()
-                    bearerTokenStorage.add(
-                        BearerTokens(accessToken = tokenInfo.accessToken, refreshToken = refreshToken),
-                    )
+                    val tokenPostResponse = client.post(tokenPostRequest)
```
Deseteral, line 120: This is fine for now, but I've noticed that `getInitialTokens` does mostly the same thing. Maybe we would like to make one method for getting the tokens.

Deseteral (approved):

PR #407, opened 2026-02-08
Implement RBAC

### LINKED ISSUE #404: Implement RBAC

gorzalczany:
Before merge:

- [ ] remove `cacheInvalidator` route which is for debug purposes only


service/src/main/kotlin/xyz/lebkuchenfm/api/RequireScope.kt
```diff
@@ -0,0 +1,28 @@
+package xyz.lebkuchenfm.api
+
+import io.ktor.http.HttpStatusCode
+import io.ktor.server.application.ApplicationCall
+import xyz.lebkuchenfm.domain.auth.Scope
+
+suspend fun ApplicationCall.missesScopes(vararg scopes: Scope): Boolean {
```
Deseteral: This method could throw, and then we could just use it without returning from controller.
So instead of this:
```kotlin
 if (call.missesScopes(Scope.XSOUNDS_UPLOAD)) {
    return@post
}
```
we would just do this instead:
```kotlin
call.requireScopes(Scope.XSOUNDS_UPLOAD)
```

service/src/main/kotlin/xyz/lebkuchenfm/api/roles/RolesEndpoint.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.ktor.server.routing.route
+import kotlinx.serialization.Serializable
+import xyz.lebkuchenfm.api.missesScopes
+import xyz.lebkuchenfm.domain.auth.Role
+import xyz.lebkuchenfm.domain.auth.Scope
+
+fun Route.rolesRouting() {
+    route("/roles") {
+        get {
+            if (call.missesScopes(Scope.USERS_MANAGE_ROLES)) {
```
Deseteral: Why should we care who can read scopes?

service/src/main/kotlin/xyz/lebkuchenfm/api/auth/ValidateAuthHandler.kt
```diff
@@ -31,6 +33,7 @@ class ValidateAuthHandler(private val authService: AuthService) {
                         call.respondWithProblem(
                             title = "Could not create new user.",
                             detail = when (error.addNewUserError) {
+                                AddNewUserError.NotFirstUser -> "Other users already exist. Contact them."
```
Deseteral: {me} When is this error created?

service/src/main/kotlin/xyz/lebkuchenfm/api/RequireScope.kt
```diff
@@ -0,0 +1,28 @@
...
+    val session = getUserSession()
+    val scopesSet = scopes.toSet()
```
Deseteral: We can just shadow this argument.
```suggestion
    val scopes = scopes.toSet()
```

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/EventStreamEndpoint.kt
```diff
@@ -21,28 +28,43 @@ fun Route.eventStreamRouting(
...
-                converter.deserialize(call.request.headers.suitableCharset(), typeInfo<EventDto>(), frame)
-            } catch (ex: SerializationException) {
-                logger.error(ex) { "Could not parse incoming WebSocket event." }
-                continue
+        eventStream.subscribe(connection)
+
+        try {
+            launch {
+                SessionInvalidationFlow.subscribe(userSession.name).collect {
+                    runCatching { close(CloseReason(CloseReason.Codes.NORMAL, "Session invalidated")) }
```
Deseteral: [runCatching is not safe to use](https://medium.com/@jatingujjar646/runcatching-with-coroutines-in-kotlin-the-hidden-cancellation-trap-92dd4ffc8d4a).

service/src/main/kotlin/xyz/lebkuchenfm/api/users/UsersEndpoint.kt
```diff
@@ -49,13 +74,89 @@ fun Route.usersRouting(usersService: UsersService) {
                     }
                 }
         }
+
+        put("/{name}/roles") {
+            if (call.missesScopes(Scope.USERS_MANAGE_ROLES)) {
+                return@put
+            }
+            val username = call.parameters["name"] ?: run {
```
Deseteral: Does this situation even occur? If `name` is missing then Ktor will never call this function (because it's not this route).

service/src/main/kotlin/xyz/lebkuchenfm/domain/auth/Scope.kt
```diff
@@ -0,0 +1,24 @@
+package xyz.lebkuchenfm.domain.auth
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+enum class Scope(val value: String) {
```
Deseteral:
```suggestion
enum class Scope(val name: String) {
```

service/src/main/kotlin/xyz/lebkuchenfm/api/eventstream/EventStreamEndpoint.kt
```diff
@@ -21,28 +28,43 @@ fun Route.eventStreamRouting(
...
-            val event = try {
-                converter.deserialize(call.request.headers.suitableCharset(), typeInfo<EventDto>(), frame)
-            } catch (ex: SerializationException) {
-                logger.error(ex) { "Could not parse incoming WebSocket event." }
-                continue
+        eventStream.subscribe(connection)
+
+        try {
+            launch {
+                SessionInvalidationFlow.subscribe(userSession.name).collect {
```
Deseteral: What if one user has more then one connection open?

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/UsersMongoRepository.kt
```diff
@@ -47,6 +55,49 @@ class UsersMongoRepository(database: MongoDatabase) : UsersRepository {
         }
     }
 
+    suspend fun migrateRoles() {
```
Deseteral: This one we need to take offline I think, since it might spark a larger discussion.
I will say this:
- this is definitely not how we should handle migrations,
- I don't think we need this specific migration.

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/UsersMongoRepository.kt
```diff
@@ -99,9 +144,30 @@ class UsersMongoRepository(database: MongoDatabase) : UsersRepository {
             }
     }
 
+    override suspend fun insertFirstUser(user: User): Result<User, InsertFirstUserError> {
+        val count = collection.countDocuments()
```
Deseteral: I wonder if we should catch exceptions from this call.

service/src/main/kotlin/xyz/lebkuchenfm/Application.kt
```diff
@@ -185,6 +207,13 @@ fun Application.module() {
             }
         }
 
+        get("/cacheInvalidator/{user}") {
```
Deseteral: This actually could be a good feature. We should change the path, method and limit it to ADMIN/OWNER scope.

Deseteral (changes requested):

PR #411, opened 2026-02-12
Commands endpoint without prompt

Deseteral:
Command prompt (`/fm`) should only be used in context where it makes sense - on Discord. Different clients would use different prompts (`fm` for terminal, `@fm` for Telegram, `!fm` in Minecraft (I guess?)).
This change makes command endpoint "prompt agnostic".

### DIFF
--- a/service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
+++ b/service/src/main/kotlin/xyz/lebkuchenfm/external/discord/DiscordClient.kt
@@ -29,7 +29,7 @@ class DiscordClient(
     private lateinit var kord: Kord
 
     private val token: String? = config.propertyOrNull("discord.token")?.getString()
-    private val commandPrompt: String? = config.propertyOrNull("commandPrompt")?.getString()
+    private val commandPrompt: String? = config.propertyOrNull("discord.commandPrompt")?.getString()
     private val channelId: String? = config.propertyOrNull("discord.channelId")?.getString()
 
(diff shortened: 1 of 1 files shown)


service/src/main/resources/application.yaml
```diff
@@ -22,5 +22,4 @@ youtube:
 discord:
   token: $?DISCORD_TOKEN
   channelId: $?DISCORD_CHANNEL_ID
-
-commandPrompt: "$COMMAND_PROMPT:/fm"
+  commandPrompt: "$COMMAND_PROMPT:/fm"
```
gorzalczany: I believe we must change reading of config in discord client now.

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/processors/HelpCommandProcessor.kt
```diff
@@ -11,7 +11,7 @@ import java.util.concurrent.ConcurrentHashMap
 
 private val logger = KotlinLogging.logger {}
 
-class HelpCommandProcessor(private val commandPrompt: String) :
+class HelpCommandProcessor :
```
gorzalczany: shouldn't we "leave" it somehow and then 
- discord would inject it as `/fm`
- "terminal" would inject it as `fm`
- minecraft would inject it as `!fm`
- etc

gorzalczany (changes requested):

PR #429, opened 2026-04-25
handle owner role assignment

gorzalczany:
What:
- refactor first user creation logic
- handle OWNER role assignement 

Why:
- in preparation for RBAC restrictions


service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/UsersMongoRepository.kt
```diff
@@ -100,6 +116,31 @@ class UsersMongoRepository(database: MongoDatabase) : UsersRepository {
             }
     }
 
+    override suspend fun insertFirstUser(user: User): Result<User, InsertFirstUserError> {
+        val count = try {
+            collection.countDocuments()
+        } catch (ex: Exception) {
+            logger.error(ex) { "An error occurred while counting users for first user insertion." }
+            return Err(InsertFirstUserError.WriteError)
```
Deseteral, line 124: Well, technically it's not a _write error_.

service/src/main/kotlin/xyz/lebkuchenfm/domain/auth/AuthService.kt
```diff
@@ -19,24 +18,27 @@ class AuthService(private val usersService: UsersService) {
...
-                Err(AuthError.UserDoesNotExistError)
+                usersService.createFirstUser(username, password)
+                    .map {
+                        logger.info { "User '$username' is the first user to log in." }
+                        UserSession(it.data.name)
+                    }
+                    .mapError {
+                        when (it) {
+                            CreateFirstUserError.UsersAlreadyExist -> {
+                                logger.info { "User '$username' tried to log in, but does not exist." }
```
Deseteral, line 30: This log message is incorrect for this error.

Deseteral (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/domain/auth/AuthService.kt
gorzalczany: well... it is :D 
- line 21:  `user == null`
- we then begin to create requested user but other userS already exist so we are not in creating mode - lets return "outer" truth - user does not exist

service/src/main/kotlin/xyz/lebkuchenfm/external/storage/mongo/repositories/UsersMongoRepository.kt
gorzalczany: yes and no

We can't write because we can't count (precondition).
I will happily change it if you have better idea :)

Deseteral: ye ye ye ye

service/src/main/kotlin/xyz/lebkuchenfm/domain/auth/AuthService.kt
Deseteral: Okay, you're right. 
But then I would argue that the fact that you had to explain this, and I had to focus to see it - does prove something ;)

Deseteral (approved):

service/src/main/kotlin/xyz/lebkuchenfm/domain/auth/AuthService.kt
gorzalczany: Agreed :) 
Maybe we should move the case "no user exists" out of auth service but not now.
Will add to backlog ;p

gorzalczany:
solves https://github.com/Deseteral/lebkuchen-fm/issues/432

PR #430, opened 2026-04-26
prepare user scopes to be handled

gorzalczany:
solves https://github.com/Deseteral/lebkuchen-fm/issues/433


service/src/test/kotlin/xyz/lebkuchenfm/api/ScopesEnforcementTest.kt
```diff
@@ -0,0 +1,121 @@
...
+            client.get("/set-session")
+            testBlock(client)
+        }
+
+    // -- withScopes tests --
+
+    @Test
+    fun `withScopes returns 200 when user has required scope`() {
+        scopesTestApp(grantedScopes = setOf(Scope.XSOUNDS_UPLOAD)) { client ->
+            val response = client.get("/with-scopes-test")
```
Deseteral: For tests I would love to see given/when/then blocks, even if they're just comments.

service/src/test/kotlin/xyz/lebkuchenfm/api/ScopesEnforcementTest.kt
```diff
@@ -0,0 +1,121 @@
...
+                    call.respond(handlerExecuted.toString())
+                }
+            }
+
+            val client = createClient { install(HttpCookies) }
+            client.get("/set-session")
+            testBlock(client)
+        }
+
+    // -- withScopes tests --
```
Deseteral: This comment is just noise.

service/src/test/kotlin/xyz/lebkuchenfm/api/ScopesEnforcementTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import xyz.lebkuchenfm.domain.auth.Scope
+import xyz.lebkuchenfm.domain.sessions.UserSession
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class ScopesEnforcementTest {
```
Deseteral: The name of this test is kind of useless I think. 
Just by looking at the name I would guess that this is integration test which checks that any given endpoint in actual application has correct scopes and they're enforced.
Instead this looks like a unit test for `withScopes` function.

service/src/test/kotlin/xyz/lebkuchenfm/api/ScopesEnforcementTest.kt
```diff
@@ -0,0 +1,121 @@
...
+                }
+
+                withScopes(Scope.XSOUNDS_UPLOAD) {
+                    get("/with-scopes-test") {
+                        handlerExecuted = true
+                        call.respondText("allowed")
+                    }
+                }
+
+                get("/handler-executed") {
```
Deseteral: Can't you make this a global static variable and check it directly in tests? Doing this indirectly via endpoint is just noise.

service/src/test/kotlin/xyz/lebkuchenfm/api/ScopesEnforcementTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    fun `withScopes returns 403 when user lacks required scope`() {
+        scopesTestApp(grantedScopes = emptySet()) { client ->
+            val response = client.get("/with-scopes-test")
+            assertEquals(HttpStatusCode.Forbidden, response.status)
+        }
+    }
+
+    @Test
+    fun `withScopes handler executes on 200`() {
```
Deseteral: This could be a simple check inside `withScopes returns 200 when user has required scope` test.

service/src/test/kotlin/xyz/lebkuchenfm/api/ScopesEnforcementTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        scopesTestApp(grantedScopes = setOf(Scope.XSOUNDS_UPLOAD)) { client ->
+            client.get("/with-scopes-test")
+            val flagResponse = client.get("/handler-executed")
+            assertTrue(flagResponse.bodyAsText().toBoolean())
+        }
+    }
+
+    @Test
+    fun `withScopes handler does not execute on 403`() {
```
Deseteral: This could be a simple check inside `withScopes returns 403 when user lacks required scope` test.

Deseteral (changes requested):

service/src/main/kotlin/xyz/lebkuchenfm/domain/users/User.kt
```diff
@@ -10,6 +11,9 @@ data class User(
 ) {
     val hasPasswordSet: Boolean get() = secret != null
 
+    val effectiveScopes: Set<Scope>
+        get() = data.roles.flatMapTo(mutableSetOf()) { it.scopes }
```
Deseteral, line 15: Later on this will become `scopes - deniedScopes`, right?

service/src/test/kotlin/xyz/lebkuchenfm/api/plugins/ScopesCheckPluginTest.kt
```diff
@@ -0,0 +1,122 @@
...
+        pluginTestApp(
+            grantedScopes = setOf(Scope.XSOUNDS_UPLOAD),
+            onHandlerExecuted = { handlerExecuted = true },
+        ) { client ->
+            // when
+            val response = client.get("/scoped-test")
+
+            // then
+            assertEquals(HttpStatusCode.OK, response.status)
+            assertEquals("allowed", response.bodyAsText())
```
Deseteral, line 74: I would even say that we don't need to check response. But this is fine.

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/ExecutionContext.kt
```diff
@@ -1,5 +1,5 @@
 package xyz.lebkuchenfm.domain.commands
 
-import xyz.lebkuchenfm.domain.sessions.UserSession
+import xyz.lebkuchenfm.domain.auth.Scope
 
-data class ExecutionContext(val session: UserSession, val commandPrompt: String?)
+data class ExecutionContext(val username: String, val grantedScopes: Set<Scope>, val commandPrompt: String?)
```
Deseteral, line 5: Why not pass whole UserSession?

Deseteral (approved):

service/src/main/kotlin/xyz/lebkuchenfm/domain/users/User.kt
gorzalczany: It was supposed to, but when I started thinking about temporary denies it became obvious that such denies won't be a simple property on the user but a separate collection in Mongo with expiry. The enforcement will then be application-level logic — something like effectiveScopes - permissionOverridesService.getDeniedScopes(user) computed at request time and cached, not baked into the user model. So effectiveScopes intentionally stays as pure role union for now.
 
I've described it in issue #436 and will implement this in later PR(s).

service/src/main/kotlin/xyz/lebkuchenfm/domain/commands/ExecutionContext.kt
gorzalczany: TBH I do not remember exact reason but it feels like unnecessary nesting to me anyway.

gorzalczany: Oh, I do remember actually. We are creating ExecutionContext in discord client from username and we were creating UserSession there and it felt wrong to me because discord does not know about "real" ktor sessions at that point. 

rephrasing: The main reason was decoupling the web session concept from the command execution context.

Issue #434, opened 2026-04-27
User roles management


gorzalczany:
Endpoints and UI for viewing and updating user roles.

**Server:**
- `PUT /api/users/{name}/roles` — accepts role name list, validates, updates
- Invalidate sessions via `SessionInvalidationFlow` on role change

**Client types & API:**
- Add `roles: string[]` to client `User` type
- Hardcode available roles (`OWNER`, `DJ`, `HONKER`, `LISTENER`) and scopes on client
- `putUserRoles(username, roles)` API function

**Users screen redesign:**
- Table with single-row selection (click to highlight)
- "New..." button → secondary `AppWindow` with create user form (moved from inline)
- "Edit..." button (disabled when no selection) → User Properties dialog

**User Properties dialog (AppWindow):**
- Username — read-only text
- Discord ID — disabled input (future: editable)
- Roles — checkboxes for OWNER, DJ, HONKER, LISTENER
- Delete password — disabled button (future: implement)
- Save / Cancel buttons

PR #437, opened 2026-04-30
Task/user roles managing

gorzalczany:
- add clients components and backend endpoints for users management

Screenshot 2026-04-30 at 21 04 36

### DIFF
--- a/client/src/apps/Users/NewUserDialog.tsx
+++ b/client/src/apps/Users/NewUserDialog.tsx
@@ -1,0 +1,60 @@
+import { AppWindow } from '@components/AppWindow/AppWindow';
+import { Input } from '@components/Input/Input';
+import { Button } from '@components/Button/Button';
+import { ErrorDialog } from '@components/ErrorDialog/ErrorDialog';
+import { USER_MANAGER_ICON_INDEX } from '@components/AppIcon/IconSpritesheet';
+import { Component, createSignal, Show } from 'solid-js';
+import { postUser } from '../../services/users-service';
+import styles from './Users.module.css';
+
+interface NewUserDialogProps {
+  close: () => void;
+  onUserCreated: () => void;
+}
+
+const NewUserDialog: Component<NewUserDialogProps> = (props) => {
+  const [error, setError] = createSignal<string | null>(null);
+
+  const onSubmit = async (e: Event) => {
+    e.preventDefault();
+    const form = e.target as HTMLFormElement;
+    const formData = new FormData(form);
+
+    try {
+      await postUser(formData);
+      props.onUserCreated();
+      props.close();
+    } catch (err) {
+      setError(err instanceof Error ? err.message : 'An unknown error occurred.');
+    }
+  };
+
+  return (
+    <>
+      <AppWindow
+        title="New User"
+        close={props.close}
+        centered
+        startSize={{ width: '300px', height: '240px' }}
+        iconIndex={USER_MANAGER_ICON_INDEX}
+      >
+        <form class={styles.dialogContent} onSubmit={onSubmit}>
+          <Input type="text" required placeholder="Username" name="username" />
+          <Input type="text" placeholder="Discord ID" name="discordId" />
+          <div class={styles.dialogButtons}>
+            <Button>Create</Button>
+            <Button type="button" onClick={() => props.close()}>
+              Cancel
+            </Button>
+          </div>
+        </form>
+      </AppWindow>
+
+      <Show when={error()}>
+        <ErrorDialog message={error()!} close={() => setError(null)} />
+      </Show>
+    </>
+  );
+};
+
+export { NewUserDialog };
--- a/client/src/apps/Users/UserPropertiesDialog.tsx
+++ b/client/src/apps/Users/UserPropertiesDialog.tsx
@@ -1,0 +1,93 @@
+import { AppWindow } from '@components/AppWindow/AppWindow';
+import { Input } from '@components/Input/Input';
+import { Button } from '@components/Button/Button';
+import { ErrorDialog } from '@components/ErrorDialog/ErrorDialog';
+import { USER_MANAGER_ICON_INDEX } from '@components/AppIcon/IconSpritesheet';
+import { Component, createSignal, For, Show } from 'solid-js';
+import { User } from '../../types/user';
+import { putUserRoles } from '../../services/users-service';
+import styles from './Users.module.css';
+
+const AVAILABLE_ROLES = ['OWNER', 'DJ', 'HONKER', 'LISTENER'];
+
+interface UserPropertiesDialogProps {
+  user: User;
+  close: () => void;
+  onSaved: () => void;
+}
+
+const UserPropertiesDialog: Component<UserPropertiesDialogProps> = (props) => {
+  const [selectedRoles, setSelectedRoles] = createSignal<string[]>([...props.user.roles]);
+  const [error, setError] = createSignal<string | null>(null);
+
+  const toggleRole = (role: string) => {
+    setSelectedRoles((prev) =>
+      prev.includes(role) ? prev.filter((r) => r !== role) : [...prev, role],
+    );
+  };
+
+  const onSave = async () => {
+    try {
+      await putUserRoles(props.user.username, selectedRoles());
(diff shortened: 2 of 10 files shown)


client/src/apps/Users/NewUserDialog.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <AppWindow
+        title="New User"
+        close={props.close}
+        centered
+        startSize={{ width: '300px', height: '240px' }}
+        iconIndex={USER_MANAGER_ICON_INDEX}
+      >
+        <form class={styles.dialogContent} onSubmit={onSubmit}>
+          <Input type="text" required placeholder="Username" name="username" />
+          <Input type="text" placeholder="Discord ID" name="discordId" />
```
gorzalczany, line 43: We should add issue for discord id editing and also think again then about endpoints restructuring.
Maybe instead of separate put roles we should just send whole User object and update discord id and roles with one "save" button click? 🤔

client/src/apps/Users/UserPropertiesDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    checked={selectedRoles().includes(role)}
+                    onChange={() => toggleRole(role)}
+                  />
+                  {role}
+                </label>
+              )}
+            </For>
+          </fieldset>
+
+          <Button disabled>Delete password</Button>
```
gorzalczany, line 75: I should probably delete it but I like the idea of showing potential missing functionalities :D 
Deleting user password would allow us for cheap "password resetting"

service/src/main/kotlin/xyz/lebkuchenfm/domain/users/UsersService.kt
```diff
@@ -143,6 +143,27 @@ class UsersService(
         return Ok(User.UserSecret(hashedPassword, salt, apiToken))
     }
 
+    suspend fun updateUserRoles(username: String, roles: Set<Role>): Result<User, UpdateUserRolesError> {
+        val user = repository.findByName(username)
+            ?: return Err(UpdateUserRolesError.UserNotFound)
+
+        if (Role.OWNER !in roles && Role.OWNER in user.data.roles) {
+            val owners = repository.findByRole(Role.OWNER)
+            if (owners.none { it.data.name != username }) {
```
Deseteral: Style: double negation if-statements are harder to read.

client/src/services/users-service.ts
```diff
@@ -27,3 +27,20 @@ export async function postUser(formData: FormData): Promise<User> {
...
+  const response = await apiFetch(`/api/users/${encodeURIComponent(username)}/roles`, {
+    method: 'PUT',
+    body: JSON.stringify({ roles }),
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+  });
+
+  if (response.ok === false) {
```
Deseteral: ???
```suggestion
  if (!response.ok) {
```

client/src/services/users-service.ts
```diff
@@ -27,3 +27,20 @@ export async function postUser(formData: FormData): Promise<User> {
...
+    throw new Error(await response.text());
```
Deseteral, line 42: I hate this, and that's ok.

client/src/components/ErrorDialog/ErrorDialog.tsx
```diff
@@ -0,0 +1,37 @@
...
+import styles from './ErrorDialog.module.css';
+
+interface ErrorDialogProps {
+  message: string;
+  close: () => void;
+}
+
+const ErrorDialog: Component<ErrorDialogProps> = (props) => {
+  return (
+    <AppWindow
```
Deseteral, line 15: Not actionable comment: I know this is because our windowing system is limited, but that's really not elegant way of doing "message box".

client/src/components/ErrorDialog/ErrorDialog.tsx
```diff
@@ -0,0 +1,37 @@
...
+      title="Error"
+      close={props.close}
+      centered
+      startSize={{ width: '300px', height: '180px' }}
+      iconIndex={USER_MANAGER_ICON_INDEX}
```
Deseteral, line 20: This component is in /components folder, but it's not really for general use. This looks like error dialog for Users app.

client/src/apps/Users/Users.module.css
```diff
@@ -38,3 +34,60 @@
...
+.usersTable > tbody > tr {
+  cursor: pointer;
+}
+
+.usersTable > tbody > tr:hover {
+  background-color: #e0e0e0;
+}
+
+.selectedRow {
+  background-color: mediumpurple !important;
```
Deseteral, line 47: Do we really need `important` here? Or did _someone_ just want a shortcut to get to the desired outcome?
🤖

client/src/apps/Users/UserPropertiesDialog.tsx
Deseteral: But now it looks like this functionality exists, but _you_ for some unknown reason cannot perform that action.

client/src/apps/Users/UserPropertiesDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Input } from '@components/Input/Input';
+import { Button } from '@components/Button/Button';
+import { ErrorDialog } from '@components/ErrorDialog/ErrorDialog';
+import { USER_MANAGER_ICON_INDEX } from '@components/AppIcon/IconSpritesheet';
+import { Component, createSignal, For, Show } from 'solid-js';
+import { User } from '../../types/user';
+import { putUserRoles } from '../../services/users-service';
+import styles from './Users.module.css';
+
+const AVAILABLE_ROLES = ['OWNER', 'DJ', 'HONKER', 'LISTENER'];
```
Deseteral, line 11: We probably shouldn't hardcode this. Especially on frontend.

client/src/apps/Users/NewUserDialog.tsx
Deseteral: I think that the whole Discord auth flow is not ideal. I think of it as a technical debt from previous LebkuchenFM.
Currently we have special command for Discord auth, which is weird since Discord is not a special case for LebkuchenFM (or at least it shouldn't be). Now that we have UI for user management it would make more sense to assign Discord ID from UI, or even better - provide a "login with Discord" button.

Deseteral (approved):

client/src/apps/Users/UserPropertiesDialog.tsx
gorzalczany: So maybe some other contributor decides to "fix" it :D

client/src/components/ErrorDialog/ErrorDialog.tsx
gorzalczany: maybe I should make it more reusable in separate PR and add error icon? or make it optional.

client/src/apps/Users/Users.module.css
gorzalczany: 🤖

client/src/apps/Users/UserPropertiesDialog.tsx
gorzalczany: I thought about it. I had endpoints to download users and scopes. But those are also hardcoded in backend. Unless we want to have runtime configurable Roles then Roles definition endpoint felt like overkill.

client/src/components/ErrorDialog/ErrorDialog.tsx
Deseteral: Sure, this sounds good.

client/src/apps/Users/UserPropertiesDialog.tsx
Deseteral: I agree that separate endpoint is overkill. And also this is fine for now.
What I had in mind is more like - what if backend service during compilation would emit JSON with all available roles, and frontend could just embed it during bundling.

Deseteral (approved):

gorzalczany:
solves https://github.com/Deseteral/lebkuchen-fm/issues/434

PR #441, opened 2026-05-15
improve app title bar

gorzalczany:
### What:
- enable phosphor icons in app title bar

### Why:
- improve its flexibility


client/src/components/AppWindow/AppWindow.tsx
```diff
@@ -19,7 +20,8 @@ interface AppWindowProps {
   title: string;
   close?: () => void;
   centered?: boolean;
-  iconIndex: IconSpriteIndex;
+  iconIndex?: IconSpriteIndex;
+  phIcon?: { type: PhIconType; icon: string };
```
Deseteral, line 24: What is `phIcon`? I know it's icon from Phosphor, but from API standpoint I wouldn't know what it will be used for.

Deseteral (changes requested):

PR #442, opened 2026-05-30
Initial biurko integration

Deseteral:


client/package.json
```diff
@@ -15,6 +15,7 @@
     "playground": "vite --config ./vite.config.playground.ts"
   },
   "dependencies": {
+    "@deseteral/biurko": "git+https://github.com/Deseteral/biurko.git#e07a930",
```
Deseteral, line 18: Let's keep pulling from git until 1.0 releases.

client/src/main.tsx
```diff
@@ -15,7 +16,7 @@ console.log(
 render(
   () => (
     <Router>
-      <Route path="/" component={Desktop} />
+      <Route path="/" component={Main} />
```
Deseteral, line 19: Had to rename because I needed another Desktop component.

client/src/views/Main/Main.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { UserAccountService } from '../../services/user-account-service';
+import { MenuBar } from '@components/MenuBar/MenuBar';
+import { SocketConnectionClient } from '../../services/socket-connection-client';
+import { PlayXSoundEventHandler } from '../../services/play-x-sound-event-handler';
+import { Desktop } from '@components/Desktop/Desktop';
+import { DesktopIcon } from '@components/DesktopIcon/DesktopIcon';
+import { useWindowManager } from '@deseteral/biurko/adapters/solid-js';
+import { ApplicationRegistry } from '../../apps/desktop-application';
+
+export function Main() {
```
Deseteral, line 11: This was `client/src/views/Desktop/Desktop.tsx` before.
`onMount` and `onCleanup` are unchanged. Rendering desktop with its icons is new.

client/src/views/Main/Main.tsx
```diff
@@ -0,0 +1,48 @@
...
+    </>
+  );
+}
+
+function DesktopApplicationList() {
+  const windowManager = useWindowManager();
+
+  return (
+    <>
+      {ApplicationRegistry.map((app) => (
```
Deseteral, line 40: I've only ported terminal application for now (not to blow up the PR).
Next PRs will convert the rest of applications to the new system.

client/package.json
gorzalczany: `"biurko": "github:Deseteral/biurko"`
and newest commit saved in .lock

client/src/components/DesktopIcon/DesktopIcon.tsx
```diff
@@ -4,19 +4,19 @@ import { IconSpriteIndex } from '@components/AppIcon/IconSpritesheet';
 
 interface DesktopIconProps {
   label: string;
-  buttonRef: (el: HTMLButtonElement) => HTMLButtonElement;
-  toggleWindow: () => void;
+  // TODO: Remove this props when migration to new system is completed.
+  buttonRef?: (el: HTMLButtonElement) => HTMLButtonElement;
+  onActivate: () => void;
```
Deseteral, line 9: `toggleWindow` -> `onActivate` since we're no longer "toggling" the view state of windows when clicking on desktop icons. Applications might as well have more then one window open at a time.

client/src/components/Desktop/Desktop.module.css
Deseteral, line 1: These are modified styles from `AppWindow`.

client/src/apps/Terminal/Buffer.tsx
```diff
@@ -1,46 +1,10 @@
...
-        </AppWindow>
-      )}
-    </>
-  );
-}
-
 const WELCOME_MESSAGE = 'Welcome to LebkuchenFM!';
 
-function Buffer() {
+export function Buffer() {
```
gorzalczany: nah, `Terminal` should stay as exported function. Buffer is internal implementation detail.
I like clean look of terminal-application though.

In this PR we should agree how do we call ApplicationWrapper vs ApplicationContent

Also what about file naming? PascalCase, camelCase, kebab-case?

client/src/apps/desktop-application.ts
```diff
@@ -0,0 +1,18 @@
+import { WindowManager } from '@deseteral/biurko';
+import { JSX } from 'solid-js';
+import { IconSpriteIndex } from '@components/AppIcon/IconSpritesheet';
+import { TerminalApplication } from './Terminal/terminal-application';
+
+type EntryPoint =
```
gorzalczany, line 10: I do not like this

client/src/apps/desktop-application.ts
```diff
@@ -0,0 +1,18 @@
...
+  | ((windowManager: WindowManager<() => JSX.Element>) => void)
+  | ((windowManager: WindowManager<() => JSX.Element>) => Promise<void>);
+
+export interface DesktopApplication {
+  name: string,
+  entryPoint: EntryPoint,
```
gorzalczany, line 16: im not a fan of this name. do we want the app to handle also other events?
eg
```
  onLaunch: (windowManager: ...) => void | Promise<void>;
  onClose?: () => void;
  onFocus?: () => void;
```

client/src/components/Desktop/Desktop.module.css
gorzalczany: I believe you
> - [x] Viewed

client/src/main.tsx
gorzalczany: now we have two main.tsx 
which one is "Mainer"?

client/src/views/Main/Main.module.css
```diff
@@ -1,9 +1,8 @@
 .desktop {
```
gorzalczany: disable scroll on that or on on the outermost container

client/src/main.tsx
Deseteral: I know. I will rename this in the cleaning PR.

client/src/apps/desktop-application.ts
Deseteral: Why?

client/package.json
Deseteral: Explained offline.

client/src/apps/Terminal/Buffer.tsx
Deseteral: Renamed Buffer to Terminal.

client/src/apps/desktop-application.ts
Deseteral: Explained offline.

client/src/views/Main/Main.module.css
Deseteral: I totally forgot about that 🤣. Fixed.

client/src/views/Main/Main.tsx
```diff
@@ -24,22 +23,24 @@ export function Main() {
...
+      <Desktop>
         <DesktopApplicationList />
       </Desktop>
     </>
   );
 }
 
 function DesktopApplicationList() {
-  const windowManager = useWindowManager();
+  const environment: SystemEnvironment = {
```
gorzalczany, line 34: could that be global in this file? I find it nested under DesktopApplicationList function rather strange :D

gorzalczany (approved):

client/src/views/Main/Main.tsx
Deseteral: No, it cannot be. Because `useWindowManager` must be used inside Solid component that is a node inside `<BiurkoDesktop>`, because that component is a context provider that contains actual `WindowManager instance (see [here](https://github.com/Deseteral/biurko/blob/main/src/adapters/solid-js/BiurkoDesktop.tsx#L74) and [here](https://github.com/Deseteral/biurko/blob/main/src/adapters/solid-js/biurko-context.ts)).

TLDR: no, because Solid.

client/src/apps/desktop-application.ts
Deseteral: Since biurko is modeled after WinAPI those "applications" only require an entry point/main function. Program lifecycles is managed by handling events.

PR #446, opened 2026-06-04
Feature: add actor and scopes to event stream

gorzalczany:
### What:
- add actor in event stream
- add xsound:listen scope

### Why:
- enable richer frontent notifications
  - eg "xyz played abc sound"
- limit who can hear xsounds


service/src/main/kotlin/xyz/lebkuchenfm/domain/eventstream/EventStream.kt
```diff
@@ -8,10 +8,13 @@ abstract class EventStream<ConsumerT : EventStream.Consumer> {
 
     abstract suspend fun sendToOne(id: EventStreamConsumerId, event: Event)
 
+    protected open fun shouldSend(consumer: ConsumerT, event: Event): Boolean = true
```
Deseteral: Maybe let's not default to `true` here?

Deseteral (dismissed):

Deseteral (approved):

